PR #52, opened 2023-07-03
Add photoionization plugins

HenningSE:
This PR will add plugins simulating delayed electrons to fuse. The source of these electrons can be photo ionization.

### DIFF
--- a/fuse/plugins/detector_physics/electron_drift.py
+++ b/fuse/plugins/detector_physics/electron_drift.py
@@ -21,7 +21,7 @@ class ElectronDrift(FuseBasePlugin):
     time and observed position is calculated.
     """
 
-    __version__ = "0.2.1"
+    __version__ = "0.3.0"
 
     depends_on = "microphysics_summary"
     provides = "drifted_electrons"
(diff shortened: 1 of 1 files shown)


ramirezdiego:
🚀

HenningSE:
Before we merge: 

- [ ] Do the tests succeed?
- [ ] Update the documentation 
- [ ] Update the plugin structure figures
- [ ] Do some manual tests and check how the PI shows up in the truth plugins

HenningSE:
Beside the failing tests, the PeakTruth plugin needs an update to account for PI peaks.

HenningSE:
Additionally we need to check the csv input part again. Looks like the updates from main were not properly merged into this branch.

HenningSE:
With the last commit, fuse will get the PI settings from the simulation config file. The corresponding PR can be found here: https://github.com/XENONnT/private_nt_aux_files/pull/306.

fuse/context.py
```diff
@@ -183,10 +212,13 @@ def set_simulation_config_file(context, config_file_name):
         for option_key, option in plugin.takes_config.items():
             if isinstance(option.default, str) and "SIMULATION_CONFIG_FILE.json" in option.default:
                 context.config[option_key] = option.default.replace(
-                    "SIMULATION_CONFIG_FILE.json",
-                    config_file_name,
+                    "SIMULATION_CONFIG_FILE.json", config_file_name
                 )
 
+            # Special case for the photoionization_modifier
```
dachengx, line 218: Why `photoionization_modifier` is special?

HenningSE: Hi @dachengx, this part is a little trick to properly set the `photoionization_modifier` config argument. The problem is that it is not written to the context config by `st.apply_xedocs_configs(version=corrections_version)` ([here](https://github.com/XENONnT/fuse/blob/0f81e47988275f866efb223bd28f0756dc7ea675/fuse/context.py#L178-L179)). As as result the `plugin.run_id` would not be overwritten by the `corrections_run_id` [here](https://github.com/XENONnT/fuse/blob/0f81e47988275f866efb223bd28f0756dc7ea675/fuse/context.py#L183-L190). 

I solved this problem by explicitly writing this config argument to the context config in the `set_simulation_config_file` function as we loop over all plugins and plugin configs anyways so we will also encounter the `photoionization_modifier` there. As it is now in the context config, `plugin.run_id` can now be changed to `corrections_run_id`.

dachengx: Will this be solved if the `photoionization_modifier` is added to the global config: https://github.com/XENONnT/corrections/blob/92aa12f5fbae460c990e04bc571af0df3f67e68d/XENONnT/global_versions/global_v14.json?

HenningSE: I think so yes. Then `st.apply_xedocs_configs(version=corrections_version)` should already take care of this.

PR #99, opened 2024-01-16
Add Truth plugins to fuse

HenningSE:
This PR will add truth output to fuse. The logic differs from the one used in WFSim so one has to be a bit careful with the interpretation of the output. 

The code in the PR works as follows:
- The cluster_id from the microphysics clustering will be propagated up to the `propagated_photons` level. For each photon, it is now possible to tell which cluster it came from.
- The `PMTResponseAndDAQ` puts these photons into waveforms and splits the waveforms into `raw_records`.  It will now save an additional output called `contributing_clusters`. This output contains the information which cluster contributed with how many photons to a given `raw_record`.
- In the next step a new `PeakTruth` plugin evaluates which clusters contribute to a `peak` and calculates truth data for each peak. One of the provided columns is the `observable_energy_truth` (we can rename columns if needed), indicating how large the energy was that went into a certain peak. Some examples:
  - Let's assume there would be a cluster of 100 keV creating 100 S2 photons. If just 95 photons make it into a peak, the observable energy for this peak will be given as 95 keV. 
  - If strax would split the S2 originating to this cluster in two peaks each with 50 photons, the observable energy for each peak would be listed as 50 keV.


mhliu0001:
Hi @HenningSE , I would say the `ap_photon_number_truth` is a bit misleading. If I understand correctly, it is actually how many photons creates afterpulses, right? But the name seems to indicate the magnitude of afterpulses in unit of PE. Maybe we should rename it, or give the magnitude (Like the sum of `photon_gain` divided by PMT gain, which doesn't need to be an integer. Maybe this is technically challenging.) instead.

dachengx:
Hey @HenningSE . Just for my more clear understanding, what is the plan for s2_tag after this PR? Is the s2_tag already in this PR?

HenningSE:
I added a draft for a cluster tagging plugin with the last commit. The plugin should evaluate if a certain cluster contributed to the main or alternative S1 or S2. Can you check if this fits your need @dachengx @mhliu0001 ?

mhliu0001:
@HenningSE Thanks for writing this draft! Yes that is close to what appletree needs. However, there are two problems here:
1. As you said before, a cluster may contribute to both main and alt S1/S2. Currently appletree separates the simulation for main/alt S2 simulation, as if they are two events. If this is the case, it doesn't make sense if appletree:
   -  either double-count the energy deposition in main S2 simulation and alt S2 simulation;
   -  or it only counts the edep in the main S2 simulation.
2. There is an upper limit to how many clusters do `contributing_clusters` contain. Although the number of clusters stored can be configured, it is hard to determine what is the correct number to account for neutron multiple scatter events.

So my suggestion is, instead of tracking the peaks, can you track the **photons** instead? For example, in `photon_summary` you have the time and the `cluster_id` that creates this photon. By time overlapping it is easy to determine whether these photons contribute to main/alt S1/S2, and it is impossible for photons to contribute to both main and alt S1/S2. After that, you can just count: N1 = how many photons are created by a certain cluster; and N2 = among these photons, how many contribute to main/alt S1/S2. Then a ratio can be stored by N2/N1.

This way, appletree can separate the main/alt S1/S2 simulation more easily, and avoid double-counting any energy deposition by multiplying the edep by the ratio.

HenningSE:
Thank you very much for your feedback @mhliu0001! I updated large parts of the PR today to follow your suggestion of using the propagated_photons directly to map the clusters to the peaks (and vice versa). Let me know what you think.

mhliu0001:
Great! I think that is what we need. I will test it with SR0 AmBe fit.

fuse/plugins/truth_information/peak_truth.py
```diff
@@ -0,0 +1,160 @@
...
+                energy_of_s1_photons_in_peak = photons_per_cluster_s1/contributing_clusters_s1["n_s1_photon_hits"] * contributing_clusters_s1["ed"]
+                #Sum up up the two: 
+                result['observable_energy_truth'][i] = np.sum(energy_of_s1_photons_in_peak) + np.sum(energy_of_s2_photons_in_peak)
+
+                result['raw_area_truth'][i] = np.sum(photons_in_peaks[i]["photon_gain"] / self.gains[photons_in_peaks[i]["channel"]])
+
+        return result
+
+        
+def get_cluster_information(interactions_in_roi, unique_contributing_clusters):
```
dachengx, line 154: If `get_cluster_information` and `weighted_position_average` are functions only used in this `.py`, we might need to add a prefix `_` to it.

fuse/plugins/truth_information/peak_truth.py
```diff
@@ -0,0 +1,160 @@
...
+
+        result = np.zeros(n_peaks, dtype=self.dtype)
+        result['time'] = peaks['time']
+        result['endtime'] = peaks['endtime']
+
+        photons_in_peaks = strax.split_by_containment(propagated_photons, peaks)
+
+        for i in range(n_peaks):
+
+            s1_photon_cut = photons_in_peaks[i]["photon_type"] == 1
```
dachengx, line 109: It might be possible to loop through S1, S2, AP, so that we do not make errors in these similar copy-pasted codes.

dachengx:
@mhliu0001 how was the "ratio" used in SR0's "S2 tagging"? There was no such a ratio, right?

HenningSE:
> In general, I understood the necessity of `ClusterTagging` as a replacement for the so-called "s2_tagging"(we might want to call it cluster tagging from now?). What are the applications of `EventTruth`, `PeakTruth`, and `RecordsTruth`. This is not an objection to these plugins, just to a better understanding.

I was thinking of studies that would need truth data like for example a new 3D position reconstruction just based on the S2 peak information. With these truth plugins you have a simple method at hand to match the positions with the peaks. Right now they are more like a "draft" showing how we could match this data. We can then later add more information to the output of these plugins if this is required for some studies and the users tell us what they need.

mhliu0001:
> @mhliu0001 how was the "ratio" used in SR0's "S2 tagging"? There was no such a ratio, right?

No, there was not. But I think I can use that when a cluster contributes to both main and alt S1/S2 (rare case, but might happen). With the current appletree framework, I can either:

 *  Split the cluster to two different clusters, one contributing to main S2 and the other contributing to alt S2, with different edep according to the ratio;
 *  Depending on the number of photons, assert that the cluster contribute to the larger of them.

fuse/plugins/truth_information/peak_truth.py
HenningSE: Makes sense! I tried to implement it, but it is not a 100% clean solution as I need to have data for S1 and S2 after the loop.

HenningSE:
I will now bump the plugin versions & update the documentation. When merging this PR we should also bump the overall fuse version to 1.1.0.

fuse/plugins/pmt_and_daq/pmt_response_and_daq.py
```diff
@@ -9,22 +9,66 @@
...
+    def iter(self, iters, executor=None):
+        return super().iter(iters, executor)
+
+    def _iter_compute(self, chunk_i, **inputs_merged):
+        return self.do_compute(chunk_i=chunk_i, **inputs_merged)
+
+    def _fix_output(self, result, start, end, _dtype=None):
+        """Wrapper around _fix_output to support the return of iterators."""
+        return result
+#############
```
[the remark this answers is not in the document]
HenningSE: The modification is not longer needed.

mhliu0001:
@HenningSE In tagged clusters, can you also include how many S1/S2 photons created by a certain cluster are detected? Just by counting the propagated_photons, should be easy. Maybe call them `photons_s1` and `photons_s2`?

dachengx (approved):

ramirezdiego (approved):

PR #121, opened 2024-02-08
PyPI build

ramirezdiego:
Adding pypi build workflow to https://pypi.org/project/xenon-fuse/.

### DIFF
--- a/.github/workflows/python_publish.yml
+++ b/.github/workflows/python_publish.yml
@@ -33,7 +33,7 @@ jobs:
     - name: Build package
       run: python -m build
     - name: Publish package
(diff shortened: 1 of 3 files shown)


HenningSE (approved):

.github/workflows/python_publish.yml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install build
+    - name: Build package
+      run: python -m build
+    - name: Publish package
+      uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
```
dachengx, line 36: I think we can be less strict on the version of `pypa/gh-action-pypi-publish`

PR #124, opened 2024-02-09
Bring the documentation to readthedocs

HenningSE:


HenningSE:
The last commit added three new very short tests that compare the plugin version number that is used in the actual code with the number that is quoted in the documentation. This test will remind us to always update the documentation when we update the plugins!

dachengx:
Hey @HenningSE . Maybe I am too late. There is a script in straxen which can build the plugins and dependency tree easily: https://github.com/XENONnT/straxen/blob/476bf1afd9dcf5015f6931c8dd1b27811ea6f1b9/docs/source/conf.py#L199. So we never needed to manually maintained straxen's plugins RTD. Maybe we can try that approach.

HenningSE:
Looks like the `TestPluginVersionsInDocumentation` tests fail as they are not able to locate the documentation source files. I don't really know why so if anyone has a good idea how to fix this I would be happy. Otherwise we need to remove these tests.

ramirezdiego:
🫠🫠🫠🫠🫠🫠🫠🫠🫠🫠🫠🫠🫠

HenningSE:
https://github.com/XENONnT/fuse/pull/124/commits/4d84b73c35ceb9c4e345afdf3404c1d54eb823ed also removed the s2_gain_spread. It is not used at the moment and we can add it back if needed in the future.

dachengx:
> Hi @dachengx, yes I saw that function but I think I would prefer a manually build documentation. It will be a bit more work when making changes but it will also be a good double-check if things make sense. Additionally it would be nice if we can expand the documentation pages in the future giving more explanations about the simulated physics and involved config arguments and outputs.

I do think this is not a good idea. In general, we should expect the users to have the ability to read the code, and we'd better see docs as the entry point of codes, not a complete description of codes. On the one hand, this will save us time in writing too many details in docs if we already written things in the docstring. On the other hand, this will save us the effort of training another maintainer of fuse.

Also, if we want to write some human language about the code, we'd better put these words in the same place, like docstring. Then the only thing we need to maintain is the docstrings, which are in the `.py` files.

dachengx:
But anyway I am not a god of code. Maybe we can ask the opinions of users of our softwares, including strax(en) about where to put the docstrings and how should we maintain the docs. @HenningSE

HenningSE:
Okay, then I would propose to keep it like it is for now and we can reconsider it in the future. I will remove the version tests now and would like to merge the PR soon so that the documentation is in a good shape.

ramirezdiego:
> Okay, then I would propose to keep it like it is for now and we can reconsider it in the future. I will remove the version tests now and would like to merge the PR soon so that the documentation is in a good shape.

I agree with this. I'd merge now the available documentation, thanks a lot for the polish!
@HenningSE Any idea of why the tests fail now?

HenningSE:
No, I postponed checking this to some later time as I don't wanted to waste a lot of time fixing this little detail.

ramirezdiego (approved):
All good now!

PR #125, opened 2024-02-10
Improvements for more straightforward and faster code

mhliu0001:
These are some minor improvements that should not alter the behavior of fuse.

### DIFF
--- a/fuse/plugins/detector_physics/s1_photon_hits.py
+++ b/fuse/plugins/detector_physics/s1_photon_hits.py
@@ -188,3 +188,7 @@ class S1PhotonHits(strax.Plugin):
         n_photon_hits = self.rng.binomial(n=n_photons, p=ly)
 
         return n_photon_hits
+
+# For here:
+# 1. get_n_photons is run multiple times, so the np.squeeze should be done when initializing the correction maps.
+# 2. s1 pattern map?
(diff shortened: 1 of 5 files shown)


fuse/plugins/detector_physics/s1_photon_hits.py
```diff
@@ -160,3 +160,7 @@ def get_n_photons(self, n_photons, positions):
         n_photon_hits = self.rng.binomial(n=n_photons, p=ly)
 
         return n_photon_hits
+
+# For here:
+# 1. get_n_photons is run multiple times, so the np.squeeze should be done when initializing the correction maps.
+# 2. s1 pattern map?
```
HenningSE, line 166: `np.squeeze` is quite fast or? At least in the tests  with 20k `ly` values (corresponding to a file of 2k Pb214 events) it took just O(500 ns). I think we can keep it that way for now.

HenningSE (changes requested):
Thanks for the PR @mhliu0001. Most of the changes look good! I would just remove the comments for the S1 photon hits, or what was your plan there?

fuse/plugins/detector_physics/s1_photon_hits.py
mhliu0001: I think I have accidentally left these comments :( These can be removed.

HenningSE (approved):
Looking good! Thanks!

dachengx:
I do not know if someone checked whether the result changed. If I change the version number back, and compare the old and new results. They are the same. So I think we are fine. Thumb to @mhliu0001 .

PR #127, opened 2024-02-14
Add missing badges and tests

ramirezdiego:
Minor polishing after #124.


HenningSE (approved):

ramirezdiego:
Thanks, @dachengx, your fix helped running the tests and we could solve a few instances where they spotted issues.

Issue #131, opened 2024-02-19
Electric field saved as integer


HenningSE:
I guess we want to have the field as float and not an integer. 

https://github.com/XENONnT/fuse/blob/0c7736a8c28325860a5d28044fe34892fc8bacbd/fuse/plugins/micro_physics/electric_field.py#L28

HenningSE:
Will be fixed with https://github.com/XENONnT/fuse/pull/133

mhliu0001:
In PR #133, the electric field is saved as float16. However, Numba does not support float16. We should instead use float32/float64. Reference: https://stackoverflow.com/questions/59447332/numba-wont-accept-numpy-array-datatype

mhliu0001:
An MWE for this problem would be: 
```python
clusters = st.get_array(run_number, ["microphysics_summary"])
events = st.get_array(run_number, "event_basics")
clusters_in_events = strax.touching_windows(clusters, events, window=10000)
```
Numba will complain about float16.

ramirezdiego:
Fixed in #135.

PR #135, opened 2024-02-20
Change electric field dtype to float32

mhliu0001:
As mentioned in issue #131 , PR #133 will cause some Numba problems because of float16 dtype. This PR changes the dtype to float32.


HenningSE (approved):
Thanks for pushing this!

ramirezdiego (approved):

Issue #136, opened 2024-02-21
custom times in csv input are ignored


KaraMelih:
It looks like the custom times are ignored when using a csv instructions
mwe
```python
import pandas as pd 
import cutax

# common folder
# custom_microphysics_instructions = pd.read_csv("/project2/lgrandi/xenonnt/simulations/supernova/sim_instructions/instructions_j4s7nrf3lw_00001.csv")
# custom_dethysics_instructions = pd.read_csv("/project2/lgrandi/xenonnt/simulations/supernova/sim_instructions/instructions_j4s7nrf3lw_00002.csv")

# check the time columns in both, they are int64, ns precision
# custom_microphysics_instructions.head(3)
# custom_dethysics_instructions.head(3)

# on fuse_context branch of cutax
context = cutax.contexts.xenonnt_fuse_full_chain_simulation()
context.set_config(
                {
                    "path": "/project2/lgrandi/xenonnt/simulations/supernova/sim_instructions/",
                    "file_name": "instructions_j4s7nrf3lw_00001.csv",
                    "n_interactions_per_chunk": 250,
                    "source_rate": 0,
                }
            )
mp = context.get_df("00001", "microphysics_summary")
```
returns all times equal to zeros. Any higher level data are not created because of this time issue.

ramirezdiego:
`"source_rate": 0` looks very suspicious to me 👀

KaraMelih:
Also, just noticed the difference between the instructed locations and the locations in microphysics summary;
The positions (`x`, `y`, `z`) are 1 order of magnitude lower. i.e. instructed position is `x=43.567`, position in microphysics summary `x=4.355` there is also a difference in the floating points but I guess that might be due to some effects.

ramirezdiego:
The positions issue could be because the input to fuse should be done in mm (as Geant4 provides the entries) and we then convert it to the cm straxen uses.

KaraMelih:
Indeed. I was not sure how `source_rate` behaves, I was under the impression that if times are manually given the source rate should be set to zero. <br>
Aaaand I found the distance unit [conversion here](https://github.com/XENONnT/fuse/blob/114143ba48daaaeb6c8e70d66dead4583e9df1a5/fuse/plugins/micro_physics/input.py#L379-L384) okkie dokkie. Thanks! Closing.

HenningSE:
[...]

PR #137, opened 2024-02-21
Fix csv timing

HenningSE:
This PR will try to solve the problems described in https://github.com/XENONnT/fuse/issues/136


KaraMelih:
Okay, checking this and the source rate change seems to be working. However, I have another (potentially silly) question; <br>
In the examples why is the radius sampled within 50cm and the z is within 150 cm? <br>
I think they both should be in millimeters right? If I run the example and look at the x,y,z distributions they are within 5cm radius and 15 cm depth. <br>

I think it does not harm, but it is potentially misleading.

ramirezdiego:
Hey @KaraMelih, from my side please go ahead and add this information to the notebook.

HenningSE:
Sounds good! Please note that I change the conversion from seconds to nanoseconds with he commit. The user now needs to provide times in nanoseconds which is easier I think.

KaraMelih:
Alright, thanks both! I set the radius to be 66cm, and z-depth to be 150 cm. I also added some comments hoping that it would clarify things. Now, I'm off to go put fuse-developer on my cv.

Issue #110, opened 2024-01-24
Automatically generate datatype and options in documentation


dachengx:
Like https://github.com/XENONnT/fuse/blob/d7b826b791749923ee24c3dc4ae68199448ed96a/docs/source/plugins/micro_physics/ChunkInput.rst, the `Technical Details`, `Provided Columns` and `Config Options` sem to be manually put into files.

Is there a possibility to automatically generate them?

dachengx:
A nice approach: https://github.com/XENONnT/fuse/pull/187

PR #138, opened 2024-02-22
Script to generate data type and kind document

zihaoxu98:
This PR aims to solve Issue #110, by auto generating the dependency tree of data types and data kinds and outputs to the readthedoc. The script is almost same as the [one](https://github.com/XENONnT/straxen/blob/master/docs/source/build_datastructure_doc.py) used in straxen. 
I dropped the data types used in straxen and only fuse data types are kept. 

I was not very careful with the texts in the added sections, so please feel free to edit them.


HenningSE:
Hi @zihaoxu98 for opening this! I think @dachengx also planed to automatize the documentation building. I would like to discuss this after we merged the Truth and PI PRs. I'm not a huge fan of these very long datastructure page, but maybe we can find a solution to split this into separate pages and unify it with the existing plugin documentation pages?

dachengx:
> Hi @zihaoxu98 thanks for opening this! I think @dachengx also planed to automatize the documentation building. I would like to discuss this after we merged the Truth and PI PRs. I'm not a huge fan of these very long datastructure page, but maybe we can find a solution to split this into separate pages and unify it with the existing plugin documentation pages?

I am very confident that we can do so. Given the long gap towards the PI PR being merged. Can we think about merging this again?

zihaoxu98:
Since the documenting is not a critical task, I can come back after other stuff is done. If there's anyone interested in picking up this before me, go ahead please.

HenningSE:
I found some free time today to work on automatic documentation updates. You can find my PR here: https://github.com/XENONnT/fuse/pull/187. I tried to stay as close as possible to the already existing documentation while getting the advantage of always up to date documentation pages. Additionally I added the dependency tree for each plugin. You can check out a preview of the documentation [here](https://xenon-fuse--187.org.readthedocs.build/en/187/).

PR #140, opened 2024-02-23
no rechunking for microphysics summary

HenningSE:


HenningSE:
Looks like our field map returned a few `nan` values that stoped nestpy from returning quanta. The last commit will set `nan` fields to 0 and print a warning message similar to the negative field values.

cfuselli (approved):
I tested it extensively and the change did solve the problems I had with the NaN values in the field. Works well now. 
Thanks @HenningSE !

ramirezdiego (commented):
Hey both, could we narrow down where in the detector these NaN pop up? Just to be sure the map handling is good and we are here addressing changes where the field is beyond our sensitive volume.

ramirezdiego:
> Hey both, could we narrow down where in the detector these NaN pop up? Just to be sure the map handling is good and we are here addressing cases where the field is beyond our sensitive volume (also to decide if setting it to zero is the right fix).

All good. @HenningSE checked the position of NaNs in our field maps and these make valid the clipping to zero.

ramirezdiego (approved):

PR #141, opened 2024-02-23
Update csv_input.py

yuema137:
A minor change
Synchronize the doc string with #137

### DIFF
--- a/fuse/plugins/detector_physics/csv_input.py
+++ b/fuse/plugins/detector_physics/csv_input.py
@@ -29,9 +29,9 @@ class ChunkCsvInput(FuseBasePlugin):
 
     source_done = False
 
-    dtype = [(("x position of the cluster [cm]", "x"), np.float32),
-             (("y position of the cluster [cm]", "y"), np.float32),
-             (("z position of the cluster [cm]", "z"), np.float32),
+    dtype = [(("x position of the cluster [mm]", "x"), np.float32),
+             (("y position of the cluster [mm]", "y"), np.float32),
+             (("z position of the cluster [mm]", "z"), np.float32),
              (("Number of photons at interaction position.", "photons"), np.int32),
              (("Number of electrons at interaction position.", "electrons"), np.int32),
              (("Number of excitons at interaction position.", "excitons"), np.int32),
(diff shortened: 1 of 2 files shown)


KaraMelih:
Hi Yue, Thanks for the PR. Indeed, the units in the docs should be consistent. However, the changes you made are on detector physics simulations input, the input for the [microphysics simulations are here](https://github.com/XENONnT/fuse/blob/e759d99d194233cdf09abd096ac8855fd9783c02/fuse/plugins/micro_physics/input.py#L32-L47) and they are wrongly labeled as `[cm]`. 

The last time I tested this, detector physics instructions accepts the positions in cm, and microphysics instructions take them in mm. Could you maybe confirm this? 

In this case, the lines you changed should actually remain "[cm]" and the lines in the micro_physics/input.py should change.

yuema137:
@KaraMelih Thanks for the comment! Please see the new update now.

HenningSE:
Hi @yuema137 and @KaraMelih, the unit [cm] [here](https://github.com/XENONnT/fuse/blob/838c072f9dcda77e2fee3a0281eb50429719cf30/fuse/plugins/micro_physics/input.py#L32-L34) and [here](https://github.com/XENONnT/fuse/blob/838c072f9dcda77e2fee3a0281eb50429719cf30/fuse/plugins/micro_physics/input.py#L215-L217) are correct. The `dtype` defines the output of the plugin, not the input. The input is in [mm] as in the Geant4 files but we directly convert them to [cm] which is used by strax(en). The conversion for root files can be found [here](https://github.com/XENONnT/fuse/blob/838c072f9dcda77e2fee3a0281eb50429719cf30/fuse/plugins/micro_physics/input.py#L381-L383) and the conversion when using csv input is [here](https://github.com/XENONnT/fuse/blob/838c072f9dcda77e2fee3a0281eb50429719cf30/fuse/plugins/micro_physics/input.py#L454-L456). 

As this is causing some confusions, we could think about adding a `fuse_input_dtype` somewhere.

KaraMelih:
Apologies for causing confusion. Yes, indeed a `fuse_input_dtype` would be nice, maybe even one for microphysics instructions and one for detector physics instructions. 

Alternatively, to avoid confusion, microphysics instructions can be made so that if a csv input is provided these are treated as cm. So the user does not have to change position inputs between detector physics and microphysics instructions, but they are rather converted under the hood

yuema137:
@KaraMelih @HenningSE Thanks for the explanation! Now it's clear to me. I'm glad to join the effort and implement the `fuse_input_dtype` as I will heavily using fuse for the pp neutrino study.

PR #142, opened 2024-02-26
Increase size of string fields in geant4_interactions

cfuselli:
Unfortunately some fields are cut out by the restriction of U<10 in the fields type, parenttype, edproc, creaproc from the root file.  I set here 18 for type and 25 for proc, I think they should cover them all (at least from what I see in my simulations). 

See for example:
type: Po214[1729.609] (15)
proc: RadioactiveDecayBase (20)


HenningSE (changes requested):
Thanks @cfuselli! Before we merge the PR there are a few more minor changes that need to be implemented. 

- [x] Bump the plugin version. (This is needed as the PR changes the dtype)
- [x] Add the changes to the [file_reader](https://github.com/XENONnT/fuse/blob/a51aaa0ada3522e46832dcc4c4bf75a017c3c655/fuse/plugins/micro_physics/input.py#L215-L230) dtype as well.
- [ ] Update the [documentation](https://github.com/XENONnT/fuse/blob/a51aaa0ada3522e46832dcc4c4bf75a017c3c655/docs/source/plugins/micro_physics/ChunkInput.rst?plain=1#L54-L71) to reflect the version bump and updated dtype.

cfuselli:
Thanks @HenningSE , added the three requested changes.

HenningSE (approved):
Looking good!

Issue #134, opened 2024-02-20
`corrections_run_id` for straxen part


mhliu0001:
For fuse part, the run_id for each plugin is substituted by the `corrections_run_id`. However, for plugins defined in `straxen`, they still use the nominal run_id instead of the `corrections_run_id`.

For example, if I use `corrections_run_id="051430"`, but try to get `["microphysics_summary", "tagged_clusters", "s1_photons", "s2_photons_sum"]` for `run_id="mc_02"`, there will be an error:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [3], in <cell line: 3>()
      1 run_number = "mc_02"
      2 # st.is_stored(run_number,"peaks")
----> 3 clusters = st.get_df(run_number, ["microphysics_summary", "tagged_clusters", "s1_photons", "s2_photons_sum"])

File ~/.local/lib/python3.9/site-packages/strax/context.py:1722, in Context.get_df(self, run_id, targets, save, max_workers, **kwargs)
   1714 def get_df(
   1715     self, run_id: ty.Union[str, tuple, list], targets, save=tuple(), max_workers=None, **kwargs
   1716 ) -> pd.DataFrame:
   1717     """Compute target for run_id and return as pandas DataFrame.
   1718 
   1719     {get_docs}
   1720 
   1721     """
-> 1722     df = self.get_array(run_id, targets, save=save, max_workers=max_workers, **kwargs)
   1723     try:
   1724         return pd.DataFrame.from_records(df)

File ~/.local/lib/python3.9/site-packages/strax/context.py:1604, in Context.get_array(self, run_id, targets, save, max_workers, **kwargs)
   1600 else:
   1601     source = self.get_iter(
   1602         run_ids[0], targets, save=save, max_workers=max_workers, **kwargs
   1603     )
-> 1604     results = [x.data for x in source]
   1606 results = np.concatenate(results)
   1607 return results

File ~/.local/lib/python3.9/site-packages/strax/context.py:1604, in <listcomp>(.0)
   1600 else:
   1601     source = self.get_iter(
   1602         run_ids[0], targets, save=save, max_workers=max_workers, **kwargs
   1603     )
-> 1604     results = [x.data for x in source]
   1606 results = np.concatenate(results)
   1607 return results

[...]

Issue #126, opened 2024-02-14
Remove unnecessary parentheses


dachengx:
The code like https://github.com/XENONnT/fuse/blob/6e98efcd3e6ed3d685d5a46fe5528b83c68ba8c6/fuse/plugins/detector_physics/s1_photon_propagation.py#L99 contains `(int)` which is equivalent to `int`. If we want to create a tuple with only `int` in it, we should use `(int,)`.

PR #143, opened 2024-02-27
Handle missing corrections_version and cleanup

HenningSE:
This PR will implement the changes discussed in https://github.com/XENONnT/fuse/issues/134 and https://github.com/XENONnT/fuse/issues/126


mhliu0001 (approved):
Looks good for me! I also checked that all unnecessary parenthesis in config types were removed.

Issue #145, opened 2024-02-27
NEST rng seed stays locked in S1 photon propagation in case there are no S1 photons


sebvetter:
During the compute step of the [S1 photon propagation plugin](https://github.com/XENONnT/fuse/blob/f5558cb856dd532de93b9b0fccf84d173ea782d8/fuse/plugins/detector_physics/s1_photon_propagation.py) the NEST rng seed is locked for the computation. It is unlocked again at the end of computation. However, if there are no S1 photons, the function exits at [line 167](https://github.com/XENONnT/fuse/blob/f5558cb856dd532de93b9b0fccf84d173ea782d8/fuse/plugins/detector_physics/s1_photon_propagation.py#L166-L167) without unlocking the seed again.

sebvetter:
Works with PR https://github.com/XENONnT/fuse/pull/146

PR #146, opened 2024-02-27
Prevent empty instructions to lock NEST random seed

sebvetter:
PR to fix https://github.com/XENONnT/fuse/issues/145


HenningSE (approved):
Thanks for fixing this! When the automatic tests are done, we can merge!

Issue #144, opened 2024-02-27
Define input format for csv inputs


HenningSE:
Based on the discussions in https://github.com/XENONnT/fuse/pull/141 we should consider adding a e.g. `fuse_input_dtype` to define the exact `dtype` needed for the csv input.

yuema137:
See #147

HenningSE:
Was done with https://github.com/XENONnT/fuse/pull/195

PR #147, opened 2024-02-28
Add fuse_input_dtype for csv inputs

yuema137:
## Add fuse_input_dtype for csv inputs

See the discussions in #141 and #144: CSV inputs are used in the `micro_physics` and `detector_physics` modules. The problem is that the requirements and units for the format needed to be explicitly defined. The main changes in this PR include the following:
### In `micro_physics.input`:
1.  Add `self.fuse_input_dtype` to the `file_loader` class in `micro_physics.input` when the `file_type` is csv
2. Check the input column names to see if it provides all of the required ones in `self.fuse_input_dtype`
3. Add docstring about the csv format to `ChunkInput` so the users can check it when using

The current check only uses `logging.warning` to check `self.column_names`. But if some columns are missing, there will be an error and it's hard for users to debug. Also, I think it's better to separate the required columns in csv from the columns we want to read from root file. So I change the column check as mentioned above.

### In `detector_physics.csv_input`:
1. Add `self.fuse_input_dtype` to the `csv_file_loader` class
2. Check the input column names to see if it provides all of the required ones in `self.fuse_input_dtype`
3. Add docstring about the csv format to `ChunkCsvInput`


yuema137:
@ramirezdiego Thanks for the comment! Sorry that I forgot to push the last commit. Now, the checks are passed.

fuse/plugins/micro_physics/input.py
```diff
@@ -209,7 +236,25 @@ def __init__(self,
...
+                      (("Track ID", "trackid"), np.int16),
+                      (("Particle type of the parent particle", "parenttype"), "<U18"),
+                      (("Trackid of the parent particle", "parentid"), np.int16),
+                      (("Geant4 process creating the particle (Redundant for csv input)", "creaproc"), "<U25"),
+                      (("Geant4 process destroying the particle (Redundant for csv input) ", 'edproc'), "<U25"),
+                      (("Event ID", "eventid"), np.int32),
+                      (("x position of the primary particle [mm]", "xp_pri"), np.float32),
+                      (("y position of the primary particle [mm]", "yp_pri"), np.float32),
+                      (("z position of the primary particle [mm]", "zp_pri"), np.float32),
+                    ]
```
ramirezdiego, line 257: From what I see, this part of the code is not very useful, since later on you anyway define `self.dtype` and perform `self.dtype = self.dtype + strax.time_fields`.

ramirezdiego (changes requested):
Hey @yuema137, thanks for the changes. I left a few comments below, but I have a couple more general comments:
1. Looking at how you defined your docstrings are functions at the input stages, this code certainly has some duplication. Notice that we did not follow the same dosctring policy for other functions.
2. Propagating here @HenningSE's suggestion: to avoid duplicated code, we could consider refactor these changes into creating a new file (e.g., `fuse_dtypes.py`) where we could define all the input dtypes for the various stages (within those, the csv microphysics input dtype). Then the user could do:
```
from fuse.fuse_dtypes import csv_microphysics_input_dtype 
input_data = np.zeros(0, dtype = csv_microphysics_input_dtype )
input_data["x"] = ....
```
to define their  input.

We can have a discussion to decide if this is an optimal format :)

fuse/plugins/micro_physics/input.py
```diff
@@ -20,7 +20,26 @@
...
+    - "zp": z position of the energy deposit [mm]
+    - "time": Time with respect to the start of the event [ns]
+        note: if the times are all zero they are distributed according to source_rate by fuse
+    - "ed": Energy deposit in keV
+    - "type": Particle type
+    - "trackid": Track ID
+    - "parenttype": Particle type of the parent particle
+    - "parentid": Trackid of the parent particle
+    - "creaproc": Geant4 process creating the particle, set for None in csv input
+    - "edproc": Geant4 process destroying the particle, set for None in csv input
```
mhliu0001, line 37: "destroying" the particle is not the correct word to use. In Geant4 that is the process in which a particle deposits energy, so a better description will be "Geant4 energy deposition process".

yuema137:
@ramirezdiego @mhliu0001 Thanks for the suggestions! I think it's a great idea to have a separate file `fuse_dtypes.py` to store all of the input dtypes and the docstrings. So that we can just import those types in `csv_input.py` and 'input.py' and remove duplicated docstring there.
Please let me know if this sounds good so I can proceed :)

ramirezdiego:
> @ramirezdiego @mhliu0001 Thanks for the suggestions! I think it's a great idea to have a separate file `fuse_dtypes.py` to store all of the input dtypes and the docstrings. So that we can just import those types in `csv_input.py` and 'input.py' and remove duplicated docstring there. Please let me know if this sounds good so I can proceed :)

Hi @yuema137, you have green light from my side. We understand that this is not a feature needed immediately, but it would be great to have it :)

PR #148, opened 2024-03-01
Change raw_records to save_when TARGET

cfuselli:
Given that for most of the analysis the raw_records are not necessary, I would like to propose to switch the plugin to strax.save_when = TARGET, this way it prevents from writing to disk a lot of data that is not needed.

Let me know if there's any reason against this change. 
I have been producing simulations like this in the past days and I find it more convenient :)


HenningSE:
Hi @cfuselli, thanks for the suggestion. I think we should discuss this with a bit broader audience and get some opinions from MC power-users as well as from computing and strax/straxen.

dachengx:
I support this. Then the users will be aware of want they are saving, especially this storage-consuming plugin.

HenningSE (approved):
Looking good! I will merge as soon as the tests are good.

Issue #151, opened 2024-03-04
Checks for interpolation methods in fuse maps


mhliu0001:
Some maps used the `RectBivariateSpline` interpolation method (like the field dependencies map), which will create unphysical values like survival probability larger than 1 or smaller than 0. See Slack thread: https://xenonnt.slack.com/archives/C0172947DFW/p1709239338483989

We should check the interpolation methods for all maps in fuse, and decide what should be used.

ramirezdiego:
@mhliu0001 Right now we use `RectBivariateSpline` for:
- field_dependencies_map
- field_distortion_comsol_map
and also in WFSim we used these methods  for the corresponding maps.

How easy is it for you to propagate your checks to all maps? We could then decide to go to `WeightedNearestNeighbors` for the next release.

mhliu0001:
I think for some maps we should use `RegularGridInterpolator`, if the data is based on a regular grid. Anyway the interpolation methods seem to be messy for all maps (see [note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:miliu:fuse_civ_interpolation) here). 

I propose to make all maps with regular bins use the `RegularGridInterpolator`, other maps should use `WeightedNearestNeighbors`. Does that make sense?

PR #152, opened 2024-03-06
Changes in interpolation methods

mhliu0001:
Currently in `straxen`, we have three interpolation methods: `RectBivariateSpline`, `RegularGridInterpolator`, `WeightedNearestNeighbors`. I propose we should avoid `RectBivariateSpline` since it creates unphysical values. The `RegularGridInterpolator` may also create unphysical values for values out of range (extrapolated), but it is faster and more straightforward than `WeightedNearestNeighbors`.

So we should use `RegularGridInterpolator` if it yields similar results to `WeightedNearestNeighbors`, and if the data are constructed on a regular bin. Otherwise we should use `WeightedNearestNeighbors`.

See [this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:miliu:fuse_civ_interpolation#interpolation_method_for_fuse_maps) for more plots to validate these changes.

Solves #151.

### DIFF
--- a/fuse/plugins/detector_physics/electron_drift.py
+++ b/fuse/plugins/detector_physics/electron_drift.py
@@ -102,7 +102,7 @@ class ElectronDrift(FuseBasePlugin):
                   'SIMULATION_CONFIG_FILE.json?'
                   '&key=field_dependencies_map'
                   '&fmt=json.gz'
-                  '&method=RectBivariateSpline',
+                  '&method=WeightedNearestNeighbors',
         cache=True,
         help='Map for the electric field dependencies',
     )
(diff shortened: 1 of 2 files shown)


mhliu0001:
I think we don't need to bump the versions, because I only changed the default configurations of the plugins. So the hash calculated by straxen now will change, won't it? In my understanding, we need to change the versions only if the hash doesn't change when the code changes.

dachengx:
> @ramirezdiego I think we don't need to bump the versions, because I only changed the default configurations of the plugins. So the hash calculated by straxen now will change, won't it? In my understanding, we need to change the versions only if the hash doesn't change when the code changes.

Hey @mhliu0001 . I do regret about suggesting you send this message. If there is nothing prevent us from updating the version, we can just do it, to be conservative and reduce the burden on understanding the policy of version updating. So please do as @ramirezdiego commented.

PR #153, opened 2024-03-06
Clean up `help` message for configs

mhliu0001:
Previously there were inconsistent `help` for the same configuration in different files. And there were also two configs (`pe_pulse_ys` and `pe_pulse_ts`) that had a placeholder description. This PR fixes these descriptions to make them consistent with each other.

Also I will let experts comment on whether my interpretation for `pe_pulse_ys` and `pe_pulse_ts` is correct.


mhliu0001:
Checking this manually is tiresome, so we should integrate this into the automatic tests.

mhliu0001:
Also in `s2_photon_propagation.py`, many `help` messages are just the same as the name of the config. I need some help here what these configs are.

HenningSE:
Hi @mhliu0001, thanks for bringing this up. For some of the variables I was not really sure how to best describe them in `help`. Many variable names were taken 1:1 from WFSim and they are not very easy to decipher. We could even think about replacing variable names like e.g. `pe_pulse_ts` with something easier to understand while collecting the informations for the help statements.

ramirezdiego:
Merged into https://github.com/XENONnT/fuse/pull/163.

PR #154, opened 2024-03-07
Fix cut_by_eventid handling

PavelKavrigin:
Same fix for `cut_by_eventid` in ROOT file loading as in the merged **epix** [PR#80](https://github.com/XENONnT/epix/pull/80).

Test notebook:
`/project2/lgrandi/pkavrigin/2024-03-07_test_fuse_cut_by/2024-03-07_TestFUSE.ipynb`
In the notebook "expected" means the number of events from G4 file which we expect to see in the epix output with the given parameters, and "result" is what we actually get.

Output with the current **fuse** master branch:

Screenshot from 2024-03-07 14-04-14

Output with `fix_cut_by_eventid` branch:

Screenshot from 2024-03-07 14-14-57


ramirezdiego:
Thanks for propagating this, @PavelKavrigin! Could you please update the version number of the plugin?
And do you know why the tests are failing?

PavelKavrigin:
I think the test time exceeds the 15 minutes limit.

PavelKavrigin:
Or maybe it's something else:
Screenshot from 2024-03-07 16-11-20

HenningSE:
Hi @PavelKavrigin thanks for bringing this to fuse! When looking through the changes I don't see any obvious reason why the tests should fail, strange. Can you maybe try to run the tests locally (`python -m unittest discover`) in the `tests` folder and see if they work there?

HenningSE:
I ran the tests locally and they still fail due to timeout errors. It seems to be related to these error messages that I get for multiple plugins:

```
TypeError: Awkward Array does not support arrays with object dtypes.

This error occurred while calling

    ak.concatenate(
        [<Array [['hadElastic', ..., 'Scintillation'], ...] type='3 * var * s...
    )
```

fuse/plugins/micro_physics/input.py
```diff
@@ -440,6 +436,13 @@ def _load_root_file(self):
         interactions["y_pri"] = ak.broadcast_arrays(xyz_pri["y_pri"], interactions["x"])[0]
         interactions["z_pri"] = ak.broadcast_arrays(xyz_pri["z_pri"], interactions["x"])[0]
 
+        if self.cut_by_eventid:
+            first_evtid = ak.min(interactions["evtid"])
+            start += first_evtid
+            stop += first_evtid
```
ramirezdiego, line 442: Hey Pavel, doesn't this part of the code affect our selection?
Let's assume we have a file with with g4_evtid == 5, and we want to process from g4_evtid == 10.

Then I set
```
entry_start == 10
cut_by_eventid == True
```
Wouldn't this line then shift it as `start = 10 + 5 = 15` ?

ramirezdiego:
Hey @PavelKavrigin, thanks a lot for checking this. We will fix the bug in https://github.com/XENONnT/fuse/pull/159, in which we decided to do this selection at the uproot loading stage (first select indices, later on load data for these indices).

PR #156, opened 2024-03-15
Prevent empty instructions from locking NEST random seed

HenningSE:


ramirezdiego (approved):

dachengx:
Just out of curiosity. Why this is necessary? @HenningSE

HenningSE:
Hi @dachengx, the problem was that if the plugin got an empty chunk as input, the nest random seed was locked and then the plugin returned an empty output without unlocking the seed. This way we could end up with some errors. With this update we never lock the random seed for empty data as the return statement if before the seed handling.

PR #159, opened 2024-03-16
Move eventid cut into uproot data loading

HenningSE:
This PR tries to move the changes from https://github.com/XENONnT/fuse/pull/154 into the data loading with uproot.

### DIFF
--- a/fuse/plugins/micro_physics/input.py
+++ b/fuse/plugins/micro_physics/input.py
@@ -132,7 +132,8 @@ class ChunkInput(FuseBasePlugin):
             n_interactions_per_chunk=self.n_interactions_per_chunk,
             arg_debug=self.debug,
             outer_cylinder=None,  # This is not running
-            kwargs={"entry_start": self.entry_start, "entry_stop": self.entry_stop},
+            entry_start=self.entry_start,
+            entry_stop=self.entry_stop,
             cut_by_eventid=self.cut_by_eventid,
             cut_nr_only=self.nr_only,
         )
@@ -184,7 +185,8 @@ class file_loader:
         chunk_delay_fraction=0.75,
         arg_debug=False,
         outer_cylinder=None,
-        kwargs={},
+        entry_start=None,
+        entry_stop=None,
         cut_by_eventid=False,
         cut_nr_only=False,
     ):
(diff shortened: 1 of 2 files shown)


ramirezdiego (approved):

fuse/plugins/micro_physics/input.py
```diff
@@ -389,29 +388,41 @@ def _load_root_file(self):
...
-        # If user specified entry start/stop we have to update number of
-        # events for source rate computation:
-        if self.kwargs["entry_start"] is not None:
-            start = self.kwargs["entry_start"]
+        # If we cut by eventid we have to read all of them first to find the start and stop index
+        if self.cut_by_eventid:
+            all_eventids = ttree.arrays("eventid")
+
+            if self.entry_start is not None:
+                start_index = np.argmax(all_eventids["eventid"] >= self.entry_start)
```
mhliu0001, line 401: I think you mean `start_index = np.argmax(all_eventids[all_eventids["eventid"] >= self.entry_start])`? Here you are performing `np.argmax` on a boolean array.

fuse/plugins/micro_physics/input.py
```diff
@@ -389,29 +388,41 @@ def _load_root_file(self):
...
+            else:
+                start_index = 0
+
+            if self.entry_stop is not None:
+                stop_index = np.argmin(all_eventids["eventid"] < self.entry_stop)
```
mhliu0001, line 406: Same problem here.

mhliu0001: Also, here you need to make sure that all_eventids[all_eventids["eventid"] < self.entry_stop] is not empty. (Same for the former one)

fuse/plugins/micro_physics/input.py
```diff
@@ -389,29 +388,41 @@ def _load_root_file(self):
...
+                if stop_index == 0:
```
mhliu0001, line 407: `stop_index == 0` can also happen when all events are in the range, in other words, `all_eventids["eventid"] < self.entry_stop` is all True.

mhliu0001: I think you mean `np.all(~(all_eventids["eventid"] < self.entry_stop))`?

PR #163, opened 2024-03-18
Cleanup and refactor configurations

ramirezdiego:
The handling of some configurations is here slightly modified, to align with the format in the fuse configuration files (see https://github.com/XENONnT/private_nt_aux_files/pull/300).

In addition, the `help` field of some configs was extended for better documentation.


fuse/plugins/detector_physics/s2_photon_propagation.py
```diff
@@ -317,19 +309,8 @@ def setup(self):
...
-            self.drift_velocity_scaling = 1.0
-            # calculating drift velocity scaling to match total drift time
-            # for R = 0 between cathode and gate
-            if "norm_drift_velocity" in self.enable_field_dependencies.keys():
-                if self.enable_field_dependencies["norm_drift_velocity"]:
-                    norm_dvel = self.field_dependencies_map_tmp(
-                        np.array([[0], [-self.tpc_length]]).T, map_name="drift_speed_map"
-                    )[0]
-                    norm_dvel *= 1e-4
-                    self.drift_velocity_scaling = self.drift_velocity_liquid / norm_dvel
```
ramirezdiego, line 332: @HenningSE Notice that I deleted this entire part, which I think was an accidental copypaste from the electron transport plugin.

ramirezdiego:
Tests are currently failing because the fuse configuration file is not yet available in our database. The simulation should run by locally sourcing the file, as @cfuselli also verified.

docs/source/plugins/detector_physics/S2PhotonPropagation.rst
```diff
@@ -141,10 +141,10 @@ S2PhotonPropagationBase plugin
      -
      - True
      - Width of the S2 area fraction top
-   * - enable_field_dependencies
+  * - enable_diffusion_transverse_map
```
mhliu0001, line 144: Missing whitespace here?

ramirezdiego: Fixed, thanks!

fuse/plugins/detector_physics/s2_photon_propagation.py
```diff
@@ -583,7 +564,6 @@ def s2_pattern_map_diffuse(self, n_electron, z, xy, drift_time_mean):
             diffusion_constant_radial *= 1e-9  # cm²/ns
             diffusion_constant_azimuthal *= 1e-9  # cm²/ns
         else:
-            # diffusion_constant_transverse = diffusion_constant_transverse
             diffusion_constant_radial = self.diffusion_constant_transverse
             diffusion_constant_azimuthal = self.diffusion_constant_transverse
```
mhliu0001, line 568: Do we need a separate configuration for azimuthal diffusion constant?

mhliu0001: Also, `enable_diffusion_transverse_map` actually means whether to use a map for both transverse and azimuthal diffusion constant, so this name is a bit misleading for me.

ramirezdiego: When we do not use the map, radial and azimuthal both become the generic transverse diffusion coefficient, hence I would say this is correct.

ramirezdiego: Before it was tricky to make the difference, but I now added the `enable_` for the booleans that decide if we call the map or not.

mhliu0001:
The error in the test is very strange. It was trying to look for `SIMULATION_CONFIG_FILE.json`, but it should be replaced by the real config file. Also, this occurs (only?) in `merge_cluster.py.`

mhliu0001:
I guess we should update the microphysics context to include `SIMULATION_CONFIG_FILE.json` replacement.

fuse/plugins/micro_physics/merge_cluster.py
```diff
@@ -62,7 +62,8 @@ class MergeCluster(FuseBasePlugin):
 
     # Config options
     tag_cluster_by = straxen.URLConfig(
-        default="energy",
+        default="take://resource://" "SIMULATION_CONFIG_FILE.json?&fmt=json" "&take=tag_cluster_by",
```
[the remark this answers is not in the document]
mhliu0001: I pushed a fix myself to debug why the automatic test failed.

mhliu0001 (approved):
Now the tests have passed, so green light from my side.

docs/source/plugins/detector_physics/S2PhotonPropagation.rst
HenningSE, line 1: We need to remove the drift_velocity_liquid config argument

HenningSE (changes requested):
Looking good to me! We need to create a SR0 config with the same format to keep backwards compatibility. 

I will request changes here to keep track of this, if we have the config in place, we are good to merge.

PR #166, opened 2024-03-19
Debug entry_stop

HenningSE:
This PR will fix the issue reported by @mhliu0001 here: https://github.com/XENONnT/fuse/pull/159#discussion_r1529325586


ramirezdiego (approved):
👍

mhliu0001:
Just to ensure, does the `entry_stop` include the edge?

ramirezdiego:
> Just to ensure, does the `entry_stop` include the edge?

No. Events up to this value should be included, excluding the value itself.

fuse/plugins/micro_physics/input.py
```diff
@@ -404,11 +404,17 @@ def _load_root_file(self):
 
             if self.entry_stop is not None:
                 stop_index = np.argmin(all_eventids["eventid"] < self.entry_stop)
```
mhliu0001, line 406: Then this line is not correct?

ramirezdiego: True 😂

PR #168, opened 2024-03-19
Update BetaYields

cfuselli:
This PR brings updates to the BetaYields plugin. 
This is an update to the plugin that was introduced in epix by @michaweiss89 and @PavelKavrigin , and that @HenningSE integrated in fuse. 

The plugin allows to get the yields for betas (nestid=8) from a specified function instead of nest. For all the other nest types, nest will be called. The excitons for betas are also sourced from nest in any case.


HenningSE:
I will close this PR in favor of https://github.com/XENONnT/fuse/pull/192. The changes are the same but the automatic tests in the other PR are not skipped.

PR #170, opened 2024-03-19
Fixing `entry_stop` selection

ramirezdiego:
... once and for all 😅 

See https://github.com/XENONnT/fuse/pull/166.
Thanks to @mhliu0001 and sorry for overlooking this.

### DIFF
--- a/fuse/plugins/micro_physics/input.py
+++ b/fuse/plugins/micro_physics/input.py
@@ -393,6 +393,13 @@ class file_loader:
             if self.entry_stop is not None:
                 log.debug(f"Ending read in at {cutby_string} {self.entry_stop}")
 
+        if self.entry_start is not None and self.entry_stop is not None:
+            if self.entry_start >= self.entry_stop:
+                 raise ValueError(
+                     "The requested range is not valid!"
+                     "Make sure that entry_stop is larger than entry_start"
+                 )
+
         # If we cut by eventid we have to read all of them first to find the start and stop index
         if self.cut_by_eventid:
             all_eventids = ttree.arrays("eventid")
--- a/fuse/plugins/micro_physics/input.py
+++ b/fuse/plugins/micro_physics/input.py
@@ -395,10 +395,10 @@ class file_loader:
 
         if self.entry_start is not None and self.entry_stop is not None:
             if self.entry_start >= self.entry_stop:
-                 raise ValueError(
-                     "The requested range is not valid!"
-                     "Make sure that entry_stop is larger than entry_start"
-                 )
+                raise ValueError(
+                    "The requested range is not valid!"
+                    "Make sure that entry_stop is larger than entry_start"
+                )
 
         # If we cut by eventid we have to read all of them first to find the start and stop index
         if self.cut_by_eventid:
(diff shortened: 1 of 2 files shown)


mhliu0001:
We should also perform a check whether `entry_stop` is larger than `entry_start`.

mhliu0001:
Another major problem is the definition of `entry_start` and `entry_stop`. Should they act on the RAW root file or after filtering out edep=0 (interactions)? Now, it is the former case, which will cause some confusion for the users.

mhliu0001:
I also added a test for this `entry_start` and `entry_stop` problem, hope that this can solve the problem once and for all.

ramirezdiego:
> Another major problem is the definition of `entry_start` and `entry_stop`. Should they act on the RAW root file or after filtering out edep=0 (interactions)? Now, it is the former case, which will cause some confusion for the users.

Hey @mhliu0001, I think this is the intended behavior. We want the user to be able to track the events processed with an index correspondance with the original ROOT file. So first we process index, later we apply the edep cut.

fuse/plugins/micro_physics/input.py
```diff
@@ -393,42 +393,58 @@ def _load_root_file(self):
...
-                        )
-                    else:
-                        log.warning("entry_stop is larger than the largest eventid in the file.")
-                        stop_index = n_simulated_events
+                if self.entry_stop <= np.min(all_eventids["eventid"]):
+                    raise ValueError(
+                        "The requested eventid range is not in the file! "
+                        "Maybe you want to set cut_by_eventid to False?"
+                    )
+                stop_index = np.searchsorted(all_eventids["eventid"], self.entry_stop)
```
ramirezdiego, line 423: Hey @mhliu0001, notice that with these last changes we almost made the same mistake 😆 `np.searchsorted` already excludes `entry_stop` once (and then a second time at the uproot level). I will change it now.

mhliu0001: No, `searchsorted` will NOT exclude `entry_stop`. You can test it with:

```
eventids = np.arange(100)
print(np.searchsorted(eventids, 99))
```

ramirezdiego: Uhm, we're confused now. @HenningSE checked and, for the previous commit, `entry_stop` and `entry_stop-1` were excluded. Now we include the `entry_stop` index, as you pointed out, which isn't what we want. Give us a minute 😁

mhliu0001: Please trust me and the automatic tests, I have (half) manually checked the root file.

HenningSE: Hi @ramirezdiego and @mhliu0001 , I tried the changes with the file used in the example notebook. If `cut_by_eventid` is set to `True`, `entry_start = 0` and `entry_stop = 10` I get the geant4 events 0 to 8. In case of `cut_by_eventid` set to `False` the same entry range gives me the events 0 to 10.

mhliu0001: Hi @HenningSE , I think this behavior is correct. If `cut_by_eventid` is `False`, it should load the first 10 entries, so evtid is `[ 0,  1,  2,  3,  4,  5,  6,  7,  8, 10]`. If `cut_by_eventid` Is `True`, it should load 0<= evtid < 10, so evtid is `[ 0,  1,  2,  3,  4,  5,  6,  7,  8]`.

HenningSE: Okay, I checked again. You are right, there is an event missing. Sorry for the confusion. It is fine like it is and we can merge!

HenningSE (approved):

PR #176, opened 2024-03-21
Exclude PMT AP photons from raw_area_truth calculation

HenningSE:


qpellegrini (approved):
It works, the issue has been fixed.

PR #178, opened 2024-03-21
Separate above- and below-gate drift of electrons

zihaoxu98:
This PR slightly modifies the drift process, to be consistent with the definition from the study shown in [this note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:zihao:fuse:sr1_s2_width_matching). This requires an additional parameter liquid_level, which is added back in https://github.com/XENONnT/private_nt_aux_files/pull/305


fuse/plugins/detector_physics/electron_drift.py
```diff
@@ -168,6 +168,12 @@ class ElectronDrift(FuseBasePlugin):
         help="Field distortion map used in fuse (Check if we can remove _fuse from the name)",
     )
 
+    liquid_level = straxen.URLConfig(
+        default="take://resource://" "SIMULATION_CONFIG_FILE.json?&fmt=json" "&take=liquid_level",
```
HenningSE, line 172: I guess the formatting was done by pre-commit but I would propose to remove the unnecessary `" "` in the string for the default value.

fuse/plugins/detector_physics/electron_drift.py
HenningSE, line 1: Please also bump the plugin version. I would suggest to go to `0.3.0`. Remember to also update the corresponding documentation file. (We should switch to an automatic documentation building soonish....)

fuse/plugins/detector_physics/electron_drift.py
```diff
@@ -327,18 +333,35 @@ def get_s2_drift_time_params(self, xy_int, z_int):
...
+            * drift_time_below_gate
+            / drift_velocity_below_gate**2
+        )
+        drift_time_spread2 += (
+            2
+            * diffusion_constant_longitudinal
+            * drift_time_above_gate
+            / drift_velocity_above_gate**2
+        )
+        drift_time_spread = np.sqrt(drift_time_spread2)
```
HenningSE, line 363: Here I would like to use a little more meaningful names for the two drift time spread values. Maybe `drift_time_spread_above_gate_squared` and `drift_time_spread_below_gate_squared` can work and then `drift_time_spread = np.sqrt(drift_time_spread_above_gate_squared + drift_time_spread_below_gate_squared)`?

HenningSE (changes requested):
Thanks for the PR @zihaoxu98! I have just some minor comments and will later test run the code.

fuse/plugins/detector_physics/electron_drift.py
zihaoxu98: Done

fuse/plugins/detector_physics/electron_drift.py
zihaoxu98: Done

fuse/plugins/detector_physics/electron_drift.py
zihaoxu98: Sure, modified

HenningSE (approved):
I will click on approve now but we should wait with merging until the config was updated.

ramirezdiego:
Thanks, @zihaoxu98! This looks good. See comments in https://github.com/XENONnT/private_nt_aux_files/pull/305 to fine-tune the configs setting for this part of the code.

ramirezdiego (approved):

dachengx:
@zihaoxu98 will this change also be applied to SR0 results? Then shall we change SR0 map?

PR #179, opened 2024-03-25
Optional random seed as user input

HenningSE:
With this PR we will give users the option to set the random seed manually. If you want to use this function, you need to set `deterministic_seed` to `False` and `user_defined_random_seed` to the seed you want to use.

### DIFF
--- a/tests/test_plugin_random_seed.py
+++ b/tests/test_plugin_random_seed.py
@@ -9,13 +9,16 @@ from _utils import test_root_file_name
 
 TIMEOUT = 60
 
+
 class TestPluginRandomSeeds(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         cls.temp_dir = tempfile.TemporaryDirectory()
         cls.run_number = "TestRun_00000"
 
-        cls.test_context = fuse.context.full_chain_context(cls.temp_dir.name, run_without_proper_corrections=True)
+        cls.test_context = fuse.context.full_chain_context(
+            cls.temp_dir.name, run_without_proper_corrections=True
+        )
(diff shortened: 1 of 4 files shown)


dachengx (changes requested):
Thanks @HenningSE . Please add a test for this feature.

HenningSE:
Thanks for the suggestion @dachengx. I added the following tests:

- test if all fuse plugins get the user defined seed
- test if all fuse plugins with a rng have a seed value defined
- test if we can intercept negative random seeds
- test if the run number changes the deterministic random seed
- test if a changed config arg changes the deterministic random seed

The last two tests are not specific to the changes in this PR but I think it does not hurt us to have them.

tests/test_plugin_random_seed.py
```diff
@@ -0,0 +1,139 @@
...
+            {
+                "deterministic_seed": False,
+                "user_defined_random_seed": -42,
+            }
+        )
+
+        # Lets check the random seed for all fuse plugins
+        for key in self.all_registered_fuse_plugins.keys():
+
+            try:
```
dachengx, line 97: Please use `with self.assertRaises(AssertionError):` but not barely `try except`.

dachengx (changes requested):
Thanks @HenningSE . All good with a small potential improvement.

tests/test_plugin_random_seed.py
HenningSE: Ahh nice! I did not know this feature yet! For plugins that have no seed I need to raise the AssertionError manually so the test will not fail.

dachengx (approved):

PR #180, opened 2024-03-27
Avoid duplicate code in downchunking

mhliu0001:
In the previous code, the `compute` function for downchunking plugins (S2PhotonPropagationBase, PMTResponseAndDAQ) has some duplicate code. While I understand this is because the chunking start/end times are different, it can be avoided by defining a function `compute_chunk`, and calling it to calculate the data in each chunk.

This PR should not alter the behavior of fuse and should be purely cosmetic. If the simulation results change, please let me know.


mhliu0001:
I don't understand why the 3.9 test passes while the 3.10 test timeout...

HenningSE (approved):
Looking good! I ran the code and the results did not change (as expected). The code in the compute method is also cleaner now. I will bump the plugin versions as we changed the compute method and then merge this PR.

PR #187, opened 2024-04-09
Automatic Plugin Documentation

HenningSE:
This PR adds a script to automatically generate the plugin documentation pages. I tried to stay close to the already existing documentation.


HenningSE:
Right now only documentation pages for plugins that are registered in the full chain context are created. We might need to tweak this step a bit to also create pages for optional plugins (e.g CSV input).

dachengx (approved):
Thanks @HenningSE . This will remove the burden of maintaining both docs and codes from analysts. I know there are still things to improve. But let's merge it now.

PR #188, opened 2024-04-10
Embellish some codes

dachengx:
1. Make the `depends_on` in the order of lower to higher plugin
2. Remove comma from `dtype`
3. Remove comma from `help`
4. Add more information in `help` but can not finish
5. straxilize `data_kind`


fuse/common.py
```diff
@@ -24,14 +24,12 @@ def dynamic_chunking(data, scale, n_min):
...
-
         elif n_cluster + 1 < n_min:
             clusters.append(c)
             n_cluster += 1
-
         elif value > scale:
-            c = c + 1
             clusters.append(c)
+            c = c + 1
             n_cluster = 0
```
HenningSE, line 33: Why was this line moved? I think we want to start a new cluster if `value > scale` and for that reason we need to bump the cluster number `c` before appending it or?

dachengx: I am wrong.

HenningSE (approved):
Looking good! Thanks @dachengx!

PR #190, opened 2024-04-15
Add lineage clustering algorithm

HenningSE:
This PR adds an alternative (micro)-clustering algorithm to fuse. The new algorithm creates clusters based on the particle track informations. It can be selected when defining the simulation context.


ramirezdiego:
Hey @HenningSE, I have executed this branch a few times in the past weeks, as well as checked the code. It looks good and I am of the opinion that, given that this approach is not yet our default, we should merge this feature as soon as it is functional. Could you then quickly summarize the changes you pushed in the past days?

HenningSE:
Hi @ramirezdiego, these changes were pushed in the last few days:

- The volume plugins, the MergeCluster and now the MergeLineage plugin all have the same dtype. I defined it in a central place so we can remove some duplicate code. 
- Add back the primary positions. They were removed earlier to simplify the dtype but we actually need them for some studies.
- Add tests running the microphysics simulation with lineage plugins registered
- In fuse main we remove all empty events and all interactions with zero energy deposit. During this PR I removed both cuts. Now I added back the empty event cut. We still save interactions with zero energy deposit as they are needed for the lineage clustering. 
- Several smaller commits to fix bugs and clean up the code.

ramirezdiego (approved):

PR #192, opened 2024-04-16
Update BetaYields

HenningSE:
This PR is a copy of the work done in https://github.com/XENONnT/fuse/pull/168. A new PR was opened to allow the automatic tests to run.


HenningSE:
Description of the PR from @cfuselli: https://github.com/XENONnT/fuse/pull/168#issue-2194743813

HenningSE:
Thanks for the nice update of the BetaYields plugin @cfuselli. There are a few changes I would like to make before we merge the PR:
- [ ] Set default values for the URLConfigs
- [ ] Move spline loading into a URLConfig
- [ ] Minor code clean up

HenningSE:
The last commit added a test for the BetaYields plugin. The test will fail as the config arguments are not properly set yet.

cfuselli:
Hey @HenningSE , @ramirezdiego, @michaweiss89  . 
On this PR I would like to propose one change: remove the g1 and g2 dependency by including it directly in the file with the spline function. I think this is better because the most important thing is that the g1 and g2 match the ones used to make the function. Do you agree? Does it make sense?

cfuselli:
@HenningSE I added a dummy function in the tests so the automatic tests finally pass. I also removed dependency from g1 and g2. The file with the function for the yields is still to be passed manually with a file path, because I think it still needs some work before uploading it. I believe the PR can go like this for the moment, to make it easier to work and test the functionality?

HenningSE:
Hi @cfuselli, the dummy function is a good idea for the tests, thanks! I think the PR is good to go and we can merge together with all the other high energy-focused PRs very soon.

cfuselli (approved):

cfuselli:
I tested it and it seems like it runs smoothly:)

PR #196, opened 2024-04-23
Scale S2 pattern map to correct S2 AFT

zihaoxu98:
This PR modifies the code of s2_aft_rescaling of s2 pattern by setting dead PMTs to be zero before rescaling, otherwise the aft could be changed even after the rescaling.


HenningSE (approved):
Looking good to me! I was thinking if we would need to bump the plugin versions to reflect the made changes in the lineage but here we change the default URLConfig to include the turned off PMTs so the lineage should be automatically updated.

ramirezdiego (approved):

Issue #199, opened 2024-04-25
Negative PMT Channel


HenningSE:
The quoted lines below can lead to negative PMT channels. This does not make sense and we should better raise an error in this case. Thanks for spotting this @Ananthu-Ravindran & @lorenzomag 

https://github.com/XENONnT/fuse/blob/fa0fff2de324c822943d684eb2718007319e0bea/fuse/plugins/detector_physics/s2_photon_propagation.py#L469-L471

PR #200, opened 2024-04-26
No photons from electrons outside

lorenzomag:
Do not return photons if all PMTs have no probability of being hit or if all electrons in a cluster are outside the TPC. In these cases, channel number is assigned negative. Only take photons with positive channel number

Solves the Issue #199 
If there are no electrons extracted within the TPC, when calculating the average position of all electrons in the cluster of extracted electrons (in this case 0), the PMT patter probability distribution is filled with not-a-number values. If this is the case, for each photon that was calculated to be emitted by the extracted electrons, the channel number is set to be -1.

If this PR is accepted, at the end of the plugin, when values are returned, all photons with negative channel numbers are rejected.

  - [x] _Does it solve one of the [GitHub open issues](https://github.com/XENONnT/fuse/issues)?_

### DIFF
--- a/fuse/common.py
+++ b/fuse/common.py
@@ -308,7 +308,6 @@ def build_photon_propagation_output(
 
     # Remove photons with photon_gain <= 0
     result = result[result["photon_gain"] > 0]
-    result = result[result["channel"] >= 0]
 
     return result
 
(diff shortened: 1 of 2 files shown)


HenningSE (approved):
Looks good!

ramirezdiego:
Ah, perfect, I see now that we only modify the S2 photon propagation plugin only :) !

PR #167, opened 2024-03-19
Lineage clustering

cfuselli:

PR #207, opened 2024-04-30
Efficient mem chunking in input plugin

cfuselli:

Changes in file_loader class of input plugin to make it more memory efficient. 
This plugin was in some cases the one using the most amount of ram during processing. 
With this new implementation, the ram requirements are reduced by at least a factor 2, and same for the speed. 
Especially needed in view of the future PR #167 when we will need to keep all interactions also with zero ed. 

Instead of transforming the full awkward array with the interactions into a numpy array immediately, we use only the times to calculate chunk boundaries, and then we will only transform into numpy array after the chunks are defined. 

This image shows the memory usage of calling next on the file_reader.output_chunk. 
The four options are: current and proposed implementation, with or without zero ed interactions. 

image

### DIFF
--- a/fuse/plugins/micro_physics/input.py
+++ b/fuse/plugins/micro_physics/input.py
@@ -242,20 +242,13 @@ class file_loader:
         if self.event_rate > 0:
             interactions["t"] = interactions["t"] - interactions["t"][:, 0]
 
-        inter_reshaped = full_array_to_numpy(interactions, self.dtype)
-
-        # Need to check start and stop again....
         if self.event_rate > 0:
             event_times = self.rng.uniform(
-                low=start / self.event_rate, high=stop / self.event_rate, size=stop - start
+            low=start / self.event_rate, high=stop / self.event_rate, size=stop - start
             ).astype(np.int64)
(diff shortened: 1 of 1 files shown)


cfuselli:
Keeps failing with:

```
=========================== short test summary info ============================
FAILED tests/test_FullChain.py::TestChunkedFullChain::test_PMTResponseAndDAQ - 
numpy.exceptions.DTypePromotionError: field titles of field 'pulse_id' mismatch
```

checked outputs and seems to be the same as before the PR, also for empty chunks it seems to give the same result. Not sure what is causing the issue here. Giving up for today..

cfuselli:
Ouch.. I thought I fixed it but it's still wrong.

cfuselli:
Fixed some mistakes, now tested again.
Based on my tests, there is no difference at all between the chunks produced before and after the PR. 
I also tested again the speed and memory usage, and updated the plot in the description. 
Results are the same, x2 memory efficiency, x2 processing speed.

ramirezdiego (approved):
Tested successfully! Faster, lighter, same chunks out :) Thanks, @cfuselli!

PR #220, opened 2024-05-15
Replace scintillation time clipping by truncation

HenningSE:

This PR will replace the s1 photon timing clipping by truncation.

Right now we are clipping the sampled S1 photon scintillation times in the S1PhotonPropagation plugin to the maximum value of `maximum_recombination_time`. To minimize the effect on the S1 waveform, this PR replaces the clipping by a truncation of the timing distribution at he same `maximum_recombination_time`. Two safety measures are added to make sure that we always have enough photon times to sample from. As a bonus the `S1PhotonPropagation.photon_timings` function was cleaned up and the scintillation timing part moved to a dedicated function.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.2.1
+current_version = 1.2.2
 commit = True
 tag = True
 
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.2.2
+current_version = 1.2.3
 commit = True
 tag = True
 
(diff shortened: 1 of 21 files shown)


HenningSE:
A note discussing the proposed changes can be found [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:analysts_overview_page:henning_schulze_eissing:s1_sc_timing_clipping).

ramirezdiego (approved):
This looks good! Like previously discussed, I prefer this re-iterative loop than assigning times for many more photons than needed, based on a safety factor. Is there a speed improvement this way?

HenningSE:
Hi @ramirezdiego, I think there is a slight performance increase using the code in this PR compared to the main branch. I did not run detailed performance tests but for the simulations done in the note above the raw records simulation time decreased by about 10%. Maybe this is not relevant to the overall simulation time when S2s are also included.

PR #221, opened 2024-05-15
Increase test timeout to 30 minutes

HenningSE:

This PR sets the test timeout (from 15 minutes) to 30 minutes.


HenningSE:
I merged the PR now as it just increases the test time-out durations. Noting controversial.

PR #225, opened 2024-05-27
Add option to distribute Geant4 events with fixed timing

HenningSE:

This PR adds the option to distribute events with a fixed timing instead of using a uniform distribution.

## Can you give a minimal working example?

Setting the tracked config argument `fixed_event_spacing` from the default value `false` to `true` will set the event timings to fixed values. The event spacing is controlled by the `source_rate` argument. 


cfuselli (approved):
I see no issues with this PR. 
Clean and simple, but very useful :) 

Both options produce a set of event times with same length and shape.

PR #226, opened 2024-05-28
Modify the definitions in peak truth to be consistent with appletree

zihaoxu98:
This PR

  - adds number of PEs in peak truth, which is simply number of photons + number of DPEs.
  - same as #210, I extend the window of peak to find photons by 200ns.
  - number of PI photons&PEs are excluded from number of S2 photons&PEs. Now the numbers of S1 S2 photons&PEs are from the scintillation only. 
  - previously, the `raw_area_truth` is from the photon gain / channel gain. However this is a random variable and does not tell us how many PEs are, because it includes the [randomly sampling from SPE spectrum](https://github.com/XENONnT/fuse/blob/7af247a8bf2342f0f4f3f134216a6bcde15d90c4/fuse/common.py#L291). Now the `raw_area_truth` is defined as the number of PEs of the corresponding scintillation type, i.e. `s1_photoelectrons_in_peak` or `s2_photoelectrons_in_peak`. Then the bias defined by `peak_area / raw_area_truth - 1` will be consistent with the definition used in [appletree](https://github.com/XENONnT/appletree/blob/ae08d50b35fdc05335664eea8f9525a7e0a405d3/appletree/plugins/reconstruction.py#L59-L61).

Since this PR highly relates to the peak reconstruction bias study, let me also tag @qpellegrini here.

  - [x] _Update the documentation_

### DIFF
--- a/fuse/plugins/truth_information/peak_truth.py
+++ b/fuse/plugins/truth_information/peak_truth.py
@@ -57,7 +57,7 @@ class PeakTruth(strax.OverlapWindowPlugin):
         ),
         (
             (
-                "Raw area from the scintillation, i.e. number of photoelectrons from the scintillation",
+                "Raw area from the scintillation, i.e. number of S1(S2) photoelectrons.",
                 "raw_area_truth",
             ),
             np.int32,
--- a/fuse/plugins/truth_information/peak_truth.py
+++ b/fuse/plugins/truth_information/peak_truth.py
@@ -27,11 +27,41 @@ class PeakTruth(strax.OverlapWindowPlugin):
         (("Number of photons from S2 scintillation in the peak.", "s2_photons_in_peak"), np.int32),
         (("Number of photons from PMT afterpulses in the peak.", "ap_photons_in_peak"), np.int32),
         (("Number of photons from photoionization in the peak.", "pi_photons_in_peak"), np.int32),
-        (("Number of photoelectrons from S1 scintillation in the peak.", "s1_photoelectrons_in_peak"), np.int32),
-        (("Number of photoelectrons from S2 scintillation in the peak.", "s2_photoelectrons_in_peak"), np.int32),
-        (("Number of photoelectrons from PMT afterpulses in the peak.", "ap_photoelectrons_in_peak"), np.int32),
-        (("Number of photoelectrons from photoionization in the peak.", "pi_photoelectrons_in_peak"), np.int32),
-        (("Raw area from the scintillation, i.e. number of photoelectrons from the scintillation", "raw_area_truth"), np.int32),
+        (
+            (
+                "Number of photoelectrons from S1 scintillation in the peak.",
+                "s1_photoelectrons_in_peak",
+            ),
+            np.int32,
+        ),
+        (
+            (
+                "Number of photoelectrons from S2 scintillation in the peak.",
+                "s2_photoelectrons_in_peak",
+            ),
+            np.int32,
+        ),
+        (
+            (
+                "Number of photoelectrons from PMT afterpulses in the peak.",
+                "ap_photoelectrons_in_peak",
+            ),
+            np.int32,
+        ),
+        (
+            (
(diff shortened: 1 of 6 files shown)


fuse/plugins/truth_information/peak_truth.py
```diff
@@ -23,11 +23,45 @@ class PeakTruth(strax.OverlapWindowPlugin):
...
+            (
+                "Number of photoelectrons from photoionization in the peak.",
+                "pi_photoelectrons_in_peak",
+            ),
+            np.int32,
+        ),
+        (
+            (
+                "Raw area from the scintillation, i.e. number of S1(S2) photoelectrons.",
+                "raw_area_truth",
```
dachengx, line 61: This PR reminds me these:

Though it is not this PR's problem, the naming `raw_area_truth` might not be most suitable. In WFSim, https://github.com/XENONnT/WFSim/blob/b75494fbd594c329eee54535b0bb9754bc7458d8/wfsim/strax_interface.py#L56 `raw_area` refers to the raw area, and is it float.

zihaoxu98: I don't know what's the most appropriate naming for this quantity. But if anyone has better idea handling this I am 100% ok.

fuse/plugins/truth_information/peak_truth.py
```diff
@@ -151,9 +213,12 @@ def compute(self, interactions_in_roi, propagated_photons, peaks):
...
 
-                    # Get the number of photons from delayed electrons
-                    photon_cut &= photons_in_peaks[i]["cluster_id"] < 0
-                    result["pi_photons_in_peak"][i] = np.sum(photon_cut)
+            # The raw area is defined by the number of PEs from the corresponding
+            # scintillation. Question: should we care mis-classification here?
+            if peak_type == 1:
+                result["raw_area_truth"][i] = result["s1_photoelectrons_in_peak"][i]
+            elif peak_type == 2:
+                result["raw_area_truth"][i] = result["s2_photoelectrons_in_peak"][i]
```
HenningSE, line 221: I think this definition is not consistent with the one needed for the peak bias study. Also, I would like to keep this independent of the `peak_type`. For the area of a peak it does not matter if the same number of photons are from the S1 or the S2 branch of the simulation.

HenningSE: Just saw that @dachengx also commented on the definition of the `raw_area`.

fuse/plugins/truth_information/peak_truth.py
```diff
@@ -86,6 +120,13 @@ class PeakTruth(strax.OverlapWindowPlugin):
         help="Drift velocity of electrons in the liquid xenon [cm/ns]",
     )
 
+    photon_finding_window = straxen.URLConfig(
+        default=200,
+        type=int,
+        help="Time window [ns] that defines whether a photon is in a peak. "
+        "Peaks' start and end times are extended by this window to find photons in them.",
+    )
+
```
HenningSE, line 129: I'm wondering if this window is maybe a bit too wide. We can have large peaks that are split in two and as a result, the end time of the first peak is the start time of the second peak. If we now have this window, we would put some photons into both peaks.

fuse/plugins/truth_information/peak_truth.py
zihaoxu98: I am asking for a change is because in appletree the bias is applied on the number of PEs (integer) to get the area (float), so my understanding is that the bias should be defined as `peak_area / number_of_PEs - 1`. If you think we should also keep raw_area's definition as it was, I can convert it back. But in the bias study I think we should change `raw_area_truth` to the number of corresponding PEs.

fuse/plugins/truth_information/peak_truth.py
zihaoxu98: You are right. It's not the same story. The left extension of peak building should be enough. Now I removed it.

dachengx (approved):

PR #228, opened 2024-05-30
Raw Records simulation acceleration in Fuse

dachengx:

The main problem in the codes before this PR:

This line https://github.com/XENONnT/fuse/blob/d5dde954301e12ebb6ec282297d98a23d47eba76/fuse/plugins/pmt_and_daq/pmt_response_and_daq.py#L380 will run `argwhere` for each `pulse["pulse_id"]`.

When the length of `unique_photon_pulse_ids` O(1e6). The running time is intolerable.

This PR first chunks the photons and then pre-processes the photons, so the first two arguments of `build_waveform`: `pulse_windows` and `photons` will have the same length. Then we do not need to run `argwhere`.

[branch name](https://jojo.fandom.com/wiki/Made_in_Heaven)

### DIFF
--- a/fuse/plugins/detector_physics/s2_photon_propagation.py
+++ b/fuse/plugins/detector_physics/s2_photon_propagation.py
@@ -338,28 +338,23 @@ class S2PhotonPropagationBase(FuseBaseDownChunkingPlugin):
 
         n_chunks = len(electron_chunks)
         if n_chunks > 1:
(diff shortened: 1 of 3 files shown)


zihaoxu98 (approved):
Thanks @dachengx This PR looks good to me! Based on my test, the PR reduces the running time significantly and the result is not changed.

PR #231, opened 2024-05-31
Keep negative gain photons

zihaoxu98:
This PR relax the requirement that photon gains must be positive. This is because for the reconstruction bias we are interested is the `peak_area / num_photoelectrons - 1`. The under-amplified photons still contribute PEs and dropping them before the truth plugin can affect the result.

After fixing this, we have consistent result with SR0 WFSim result. Check [the notebook](https://github.com/XENONnT/analysiscode/blob/master/SR1_fuse_matching/peak_recon_bias.ipynb) for details.

### DIFF
--- a/fuse/common.py
+++ b/fuse/common.py
@@ -323,9 +323,6 @@ def build_photon_propagation_output(
     result["cluster_id"] = _cluster_id
     result["photon_type"] = photon_type
 
-    # Remove photons with photon_gain <= 0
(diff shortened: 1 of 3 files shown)


FaroutYLq (approved):

FaroutYLq:
This makes a lot more sense to me physically, but I would let core maintainer decide whether to merge.

ramirezdiego (approved):

PR #235, opened 2024-06-07
Overriding NR Scintillation Delay Time if Specified

FaroutYLq:

Cooperating with [this PR](https://github.com/XENONnT/private_nt_aux_files/pull/319). If specified scintillation delay override and we want to simulate NR, then we will override the scintillation time [here](https://github.com/XENONnT/fuse/blob/c2baf86e9e85b477f050acc97e013feb69d43593/fuse/plugins/detector_physics/s1_photon_propagation.py#L378). This is to achieve better matching between NR calibration data and simulation. No notes yet given the time pressure but I will write one and link here. For now see discussion [here](https://xenonnt.slack.com/archives/C0172947DFW/p1717613957524749) for reference.

### DIFF
--- a/fuse/plugins/detector_physics/s1_photon_propagation.py
+++ b/fuse/plugins/detector_physics/s1_photon_propagation.py
@@ -311,7 +311,8 @@ class S1PhotonPropagation(S1PhotonPropagationBase):
         "&take=s1_nr_scint_time_ed_override",
         type=(int, float),
         cache=True,
-        help="This will only be used in NR scintillation dealy time. In keV unit. Overriding the energy of NEST scintillation delay model for better match to NR data.",
+        help="This will only be used in NR scintillation dealy time. In keV unit."
+        "Overriding the energy of NEST scintillation delay model for better match to NR data.",
     )
 
     s1_nr_scint_time_nesttype_override = straxen.URLConfig(
@@ -320,7 +321,8 @@ class S1PhotonPropagation(S1PhotonPropagationBase):
         "&take=s1_nr_scint_time_nesttype_override",
         type=(int, float),
         cache=True,
-        help="This will only be used in NR scintillation dealy time. Overriding the NESTtype of NEST scintillation delay model for better match to NR data.",
+        help="This will only be used in NR scintillation dealy time."
+        "Overriding the NESTtype of NEST scintillation delay model for better match to NR data.",
     )
 
     s1_nr_scint_time_excitonfrac_override = straxen.URLConfig(
@@ -329,7 +331,8 @@ class S1PhotonPropagation(S1PhotonPropagationBase):
         "&take=s1_nr_scint_time_excitonfrac_override",
         type=(int, float),
         cache=True,
-        help="This will only be used in NR scintillation dealy time. Overriding the exciton fraction of NEST scintillation delay model for better match to NR data.",
+        help="This will only be used in NR scintillation dealy time. Overriding the exciton" 
+        "fraction of NEST scintillation delay model for better match to NR data.",
     )
 
     def setup(self):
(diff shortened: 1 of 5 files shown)


fuse/plugins/detector_physics/s1_photon_propagation.py
```diff
@@ -374,14 +410,23 @@ def nest_scintillation_timing(
...
-
-                scint_time = self.nestpy_calc.GetPhotonTimes(
-                    nestpy.INTERACTION_TYPE(recoil_type[i]),
-                    n_times_to_sample,
-                    n_excitons[i],
-                    local_field[i],
-                    e_dep[i],
-                )
+                # If NR types, we check if we want to use the effective scintillation delay model
+                if self.override_s1_nr_scint_time and (recoil_type[i] <= 6):
```
FaroutYLq, line 414: @ramirezdiego please take a look if this makes sense to you. I was following 
```
    nr=0, wimp=1, b8=2, dd=3, ambe=4, cf=5, ion=6, gammaray=7,
    beta=8, ch3t=9, c14=10, kr83m=11, nonetype=12
```
from my old knowledge of NEST types

FaroutYLq:
Once we switched to explicit default, we [passed auto tests](https://github.com/XENONnT/fuse/pull/235/commits/b7d2e684f9c1d5d26fdb9da508f8217449e63914). I would suggest we ignore the lockseed issues, since local tests can be passed.

FaroutYLq:
Once we got green light from @ramirezdiego I think we can merge this one.

FaroutYLq:
@dachengx nicely figured out why the test failed because of seed lock. I would invite him to explain here for future records.

dachengx:
> @dachengx nicely figured out why the test failed because of seed lock. I would invite him to explain here for future records.

Background knowledge: pytest uses a single process and multi-threads.

For a plugin like `S1PhotonPropagation` that call `lock_seed` and `unlock_seed` inside its `compute` method, if an error happens after `lock_seed` and before `unlock_seed`, the `nest_rng` random generator will not be unlocked. So that other threads will see an error when calling `lock_seed`:
```
RuntimeError: You can not change the seed because it is locked.
```

ramirezdiego:
I don't get why this is still failing. The nestpy version locked should be the correct one.

dachengx:
> I don't get why this is still failing. The nestpy version locked should be the correct one.

After the new .json configuration file is available, the error will be gone. Reason: https://github.com/XENONnT/fuse/pull/235#issuecomment-2156301936

ramirezdiego:
> > I don't get why this is still failing. The nestpy version locked should be the correct one.
> 
> After the new .json configuration file is available, the error will be gone. Reason: [#235 (comment)](https://github.com/XENONnT/fuse/pull/235#issuecomment-2156301936)

Hey @dachengx, I get the error and the solution for it, but then I do not get why we had to lock the nestpy version -- this is what confuses me. We were probably using the right version from the beginning, weren't we?

dachengx:
> > > I don't get why this is still failing. The nestpy version locked should be the correct one.
> > 
> > 
> > After the new .json configuration file is available, the error will be gone. Reason: [#235 (comment)](https://github.com/XENONnT/fuse/pull/235#issuecomment-2156301936)
> 
> Hey @dachengx, I get the error and the solution for it, but then I do not get why we had to lock the nestpy version -- this is what confuses me. We were probably using the right version from the beginning, weren't we?

It is changed back.

ramirezdiego:
For completeness on what just happened: the error indeed came from a lock seed after a code is executed and fails (as reported above by @dachengx). The confusion that I wanted to clarify is that the solution is not to lock the nestpy version (which I also like, and we do in our containers (https://github.com/XENONnT/montecarlo_environment/blob/43f4bf62d6e7a883bacb6ac80c645a54f235fdc5/create-env#L98), but to make all the  configs for this PR available, such that the function runs and the random seed issue does not pop up. Tests running now, thanks @FaroutYLq and @dachengx!!

PR #236, opened 2024-06-08
Duplicate PR 235 to see if we can pass tests

FaroutYLq:

Duplicate #235 to see if we can pass tests. 

Cooperating with [this PR](https://github.com/XENONnT/private_nt_aux_files/pull/319). If specified scintillation delay override and we want to simulate NR, then we will override the scintillation time [here](https://github.com/XENONnT/fuse/blob/c2baf86e9e85b477f050acc97e013feb69d43593/fuse/plugins/detector_physics/s1_photon_propagation.py#L378). This is to achieve better matching between NR calibration data and simulation. No notes yet given the time pressure but I will write one and link here. For now see discussion [here](https://xenonnt.slack.com/archives/C0172947DFW/p1717613957524749) for reference.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -47,6 +47,7 @@ jobs:
           git clone https://github.com/NESTCollaboration/nestpy.git
           cd nestpy
           git submodule update --init --recursive
+          git checkout fb3804e
           pip install .
           cd ..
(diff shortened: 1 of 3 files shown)


FaroutYLq:
No it doesn't work. We fail still

PR #241, opened 2024-06-12
Keep interactions even if outside NEST validity

cfuselli:
Instead of removing the interaction, keep it and just give the warning. 

I find the "Remove interaction" much more dangerous than using a yield model that is not completely validated by data from NEST. The function can return values also for energies much above the validity range, and seems to give reasonable results. 

Instead, removing interactions can result into tricky results if the analyst is not fully aware of the possibility. 

Here attached the model validity and the results of  nc.GetYields() for energy ranges outside the validity ( 3MeV for gamma and beta, 200keV for NR). 

test_fuse_yields_nest_nr_yields
test_fuse_yields_nest_gamma_yields
test_fuse_yields_nest_beta_yields

Just to make it extra clear: 

- the smooth solid lines are the output of nestpy.PhotonYield() and nestpy.ElectronYield(), they give NaN outside the validity range.
- the lines that oscillate more are the output of nc.GetYields().getQuanta() - that is what is used in the plugin, and gives values also outside the validity range. 

The functions that define the yields model come from here in [NEST](https://github.com/NESTCollaboration/nest/blob/2259ec4cd56029033409d92ff60246629bca124a/src/NEST.cpp#L1016).


HenningSE (approved):
Thanks, @cfuselli I agree: Just removing these energy deposits is dangerous and will harm us more than having a questionable yield model. I would propose merging this PR with the other high-energy-focused PRs.

HenningSE:
Ohh and one more thing: Can you bump the plugin version?

ramirezdiego (approved):
Forgot to approve this after discussing with Carlo 😆 Thanks, Henning!

PR #245, opened 2024-07-03
Update yields and clustering

cfuselli:
### Refactor NestYields
The plugin is refactored to be more flexible and easier to replace single functions in alternative plugins like BetaYields. 
Now we define a NestCalc object in the setup to use it in different functions. We vectorise get_quanta instead of get_quanta_from_nest and we separate yields and quanta calculations into two different functions. 
The content of the plugin itself and the results are unchanged. 

### BetaYield modifications
The plugin now intervenes only on the results of the yields from nest, but still makes use of the getQuanta function from NEST where all the statistical fluctuations ( fano factors etc ) and recombination fluctuations are calculated. 

### New option: tune NEST width parameters ( NestYields ) 
We now allow to specify the NRERWidthsParameter that are used by nest to do all the calculations on recombination fluctuation etc, to go from pure yield to actual number of quanta produced. These parameters can be tuned to match our data. Leave dict empty to use defaults from NEST. 

### New option: return directly yields model output
We add an option to produce simulations turning off all statistical fluctuations and recombination fluctuation. 
Handy to study response decoupling from other effects. Use with the option `return_yields_only`.

### New option: Beta yields threshold
Introduce an energy threshold for using the BetaYield ( like only >10keV ) - it's difficult to make a function that goes well also there, and nest gives good behavior at very low energies, setting manually the yields model to 0 at 0.3keV for LY and 0.01keV for CY. This is necessary for us to avoid having yield model going to infinity at these very low values. Use with the option `beta_yields_threshold`, default to 10keV.

### New option: Fix electric field for yields ( BetaYields ) 
We can now decide to fix the value of electric field that is used to calculate the yields, separately for gammas and betas, using `fix_beta_yield_field` and `fix_gamma_yield_field`. 

### Improvement: Lineage clustering logic
We now add the case of a gamma-phot interaction happening inside a photo-absorption (also gamma-phot) cluster. This additional interactions could be due to x rays emitted. 

### Bug fix: Lineage cluster classify
It does not matter what the creaproc was, if it is gamma-phot, give gamma yields. 

### Bug fix: no electrons if create_S2 is False
In the implementation of BetaYields, the treatment of create_S2 was missing. Now we correctly set to zero the production of electrons in the plugin if the create_S2 of the interaction is set to False. 

### Removal: recombination fluctuation model
The current implementation does not support anymore an empirical data motivated model for recombination fluctuation, as this is now done with all the necessary details by the nest getQuanta function.


fuse/plugins/micro_physics/yields.py
```diff
@@ -33,6 +33,12 @@ class NestYields(FuseBasePlugin):
 
     save_when = strax.SaveWhen.TARGET
 
+    use_recombination_fluctuation = straxen.URLConfig(
+        default=True,
+        type=bool,
+        help="Turn on or off the recombination fluctuation for beta interactions.",
+    )
```
HenningSE, line 40: Hi @cfuselli, from looking at the code, `use_recombination_fluctuation` will also work for our pure NEST quanta calculation or? I would change the help message and just call it something like: `"Turn on or off the recombination fluctuation"`

fuse/plugins/micro_physics/yields.py
```diff
@@ -174,75 +167,57 @@ class BetaYields(NestYields):
...
-        # now for the beta interactions we use the beta yields
-        photons_beta, electrons_beta = self.quanta_from_spline(
-            interactions_in_roi["ed"][mask_beta],
-        )
+        if self.beta_quanta_spline is None:
+            raise ValueError("beta_quanta_spline must be set in the context config")
 
-        photons[mask_beta] = photons_beta
-        electrons[mask_beta] = electrons_beta
+        super().setup()
```
HenningSE, line 186: Do we need this explicit check if the spline is set? I guess strax will give us an error message anyway or?

HenningSE (commented):
Not much to add, the code is very clean. I will try to test is and if it runs fine I will approve.

fuse/plugins/micro_physics/lineage_cluster.py
```diff
@@ -407,6 +400,12 @@ def is_lineage_broken(
 
         distance = np.sqrt(np.sum((parent_position - particle_position) ** 2, axis=0))
 
+        if (particle["creaproc"] == "phot") and (particle["edproc"] == "phot"):
```
cfuselli, line 403: This was necessary to make sure that we do not split a photoionisation peak into multiple parts. This is because there can be some additional ~35keV gamma-phot inside the photoionisation peak ( probably from x-rays )

ramirezdiego:
Ciao @cfuselli, when you find some time, could you update the topmost description of your PR to account for the recent changes and the merged parallel branch? Thanks for the good work!

HenningSE (approved):
Very nice PR, thank you @cfuselli and @michaweiss89! I would like to wait for @ramirezdiego to have a last look and then we can merge

PR #246, opened 2024-07-11
Update mass and charge dtypes to int16

cfuselli:
Because the maximum for int8 is 127, we get overflows when representing for example A=212 ( goes to -44). 
Moving to int16 should have no big impact and will solve the issue.


HenningSE (approved):
Thanks! Looks good!

Issue #247, opened 2024-07-12
Log debug not printing


cfuselli:
I think we should change log.debug to self.log.debug in many locations inside the plugins. Otherwise setting fuse to debug mode will not print the debug statements, as the debug mode is set for the self.log of the plugin.

PR #248, opened 2024-07-15
Fix plugin warning/debug/info logging

HenningSE:

This PR aims at fixing logging messages that did not show up during the simulation. The problem is described here: https://github.com/XENONnT/fuse/issues/247

  - [x] _Does it solve one of the [GitHub open issues](https://github.com/XENONnT/fuse/issues)?_

### DIFF
--- a/fuse/plugins/micro_physics/yields.py
+++ b/fuse/plugins/micro_physics/yields.py
@@ -201,6 +201,7 @@ class BetaYields(NestYields):
             interactions_in_roi["A"],
             interactions_in_roi["Z"],
             interactions_in_roi["create_S2"],
+            log=self.log,
             density=interactions_in_roi["xe_density"],
         )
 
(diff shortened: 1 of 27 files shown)


fuse/plugins/micro_physics/yields.py
```diff
@@ -87,13 +83,14 @@ def get_quanta(self, interactions_in_roi):
...
             interactions_in_roi["create_S2"],
+            log=self.log,
             density=interactions_in_roi["xe_density"],
         )
 
         return photons, electrons, excitons
 
     @staticmethod
-    def _quanta_from_NEST(en, model, e_field, A, Z, create_s2, **kwargs):
+    def _quanta_from_NEST(en, model, e_field, A, Z, create_s2, log, **kwargs):
```
cfuselli, line 93: this will change with PR #245 , but in principle it is good. (this function will not be static anymore, so self.log will be good)

cfuselli (approved):
looks good, thank you @HenningSE !!

fuse/plugins/micro_physics/yields.py
HenningSE: Hi @cfuselli you are right, I will update this part when resolving the merge conflicts of the linked PR.

Issue #239, opened 2024-06-09
Exciton Misplaced?


FaroutYLq:
### What's the problem?

It seems to me that we might be doing something unphysical regarding treatment of excitons.

1. Excitons are first computed [here](https://github.com/XENONnT/fuse/blob/75429f426376d93276de8c1039ab9b18acdc8c18/fuse/plugins/micro_physics/yields.py#L66) or [there](https://github.com/XENONnT/fuse/blob/75429f426376d93276de8c1039ab9b18acdc8c18/fuse/plugins/micro_physics/yields.py#L322), together with photons and electrons **before** photon/electron collection efficiency applied. That means, the excitons we got here are also **before** photon/electron collection efficiency applied.
2. The next and also last time excitons show up in `fuse`, is in S1 photon propagation. It was called in computing[ scintillation delay timing](https://github.com/XENONnT/fuse/blob/75429f426376d93276de8c1039ab9b18acdc8c18/fuse/plugins/detector_physics/s1_photon_propagation.py#L378-L384). Together with it in the same function `nestpy_calc.GetPhotonTimes`, the corresponding photons are already **after** collection efficiency. 

Note that, we never implement a binomial process representing the loss of those excitons' photons in `s1_photon_hits` [here](https://github.com/XENONnT/fuse/blob/75429f426376d93276de8c1039ab9b18acdc8c18/fuse/plugins/detector_physics/s1_photon_hits.py#L166). That being said, it is not surprising that you will have a lot of events with actual exciton fraction greater than 1, which is completely unphysical and already enters the wild extrapolation region in NEST. (Such breaking of physical sanity is even allowed in NEST!! See my PR for NEST [here](https://github.com/NESTCollaboration/nest/pull/186)). 

### What it affects?
Of course, all the S1 pulse shape dependent studies are affected. However, as long as we have validated our simulation with data and got acceptable matches, we are not really wrong. There is no reason to interpret all the processes from NEST as physical, and it is OK we just use it as a grey box, with unphysical parameters fed into it.

### How to do better?
A fix will be straightforward, just instead of only considering the [total photons](https://github.com/XENONnT/fuse/blob/75429f426376d93276de8c1039ab9b18acdc8c18/fuse/plugins/detector_physics/s1_photon_hits.py#L166), apply two binomial processes and then find a way to return both photons and excitons:
```
n_recombined = n_photons - n_excitons
n_photon_hits_excitons = self.rng.binomial(n= n_excitons, p=ly)
[...]

PR #249, opened 2024-07-16
Exciton ratio

HenningSE:

This PR will exchange the number of excitons in the S1 timing simulation by the ratio of excitons to photons. It aims at solving https://github.com/XENONnT/fuse/issues/239. 

### DIFF
--- a/fuse/dtypes.py
+++ b/fuse/dtypes.py
@@ -61,6 +61,7 @@ quanta_fields = [
     (("Number of photons at interaction position", "photons"), np.int32),
     (("Number of electrons at interaction position", "electrons"), np.int32),
     (("Number of excitons at interaction position", "excitons"), np.int32),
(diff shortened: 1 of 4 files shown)


ramirezdiego:
Thanks, Henning! To recap, we should agree on one of these two options:

1. We calculate this ratio in the yields plugin, but then need to warn the user to set it in the instructions that bypass this step (and modify the csv input instructions accordingly).
2. We make the calculation of this ratio only within the S1 propagation class.

I prefer the second, to avoid burdens to the users, but either is fine.

HenningSE:
Hi @ramirezdiego, thanks for your feedback. I opened a PR with option 2 here https://github.com/XENONnT/fuse/pull/251. We can then decide which one is better and merge one of the two PRs.

ramirezdiego:
Thanks! I prefer the second, hence closing this one.

PR #254, opened 2024-08-11
Fix lineage algo - RadioactiveDecayBase

cfuselli:
With this PR we fix some issues with the treatment of RadioactiveDecayBase (RDB) interactions. I realised that the results of simulating one gamma and simulating the gamma from Xe131 gave very different results. That is not good. Now they both give similar results. 

Specifically, we had to add special treatment for gammas emitted from excited nuclei that are treated by RDB library. 
Low energy gammas will not produce a gamma-phot/compt interaction in G4 just like hig energy gammas, but just some electron tracks and some gamma tracks will be produced. Now, we decide based on the secondaries what yield to assign to a certain cluster. The modifications match perfectly the distributions of Xe129 and Xe133 ( they were completely off before this PR ) and also match much better Pb212. 

## Details of changes: 

- If there is "[" in the particle type ( like Xe129[236.140] ) - we pass also the secondaries to the classification. We need them to understand if to give a gamma or beta yield.
- Added get_particle_secondaries function to get first order daughters of a particle
- classification with secondaries: if there are only gammas -> it's a beta yield compton. If there are also electrons, it's probably a low energy gamma, so gets a gamma yield. 
- lineage broken: brake if creaproc and edproc are both RDB. Intermediate step. 
- modify ion lineage broken. before: brake if lineage is 6, now checks if ion. 
- added: if lineage is a gamma yield, never brake anymore. We don't want to split photoabsorb. 
- added: 0.1cm threshold specially for brem. 

#### Xe129 - 39keV gamma
It makes some electron tracks and some gamma-phots. We want to keep them all together into a one gamma cluster. 

Xe129_39kev

#### Xe129 - 196keV gamma
It makes some electron tracks. Again, we want to keep them all together into a one gamma cluster. 

Xe129_196kev

#### Bi214 - 609keV gamma
It makes a real gamma. The gamma produces two compton clusters ( beta yield ) and one photoabsorbtion ( gamma yield ) 

BiPo_609kev

 ## Summary of clustering for different situations

Left: NEW clustering 
Right: OLD clustering

#### Gamma simulations 163keV in TPC
Expect just one gamma cluster, sometimes compton + gamma

Gamma_summary

#### Pb212
It's gammas+beta most of the times, with 238keV gamma

Pb212_summary

# Xe131
It's one 163keV gamma
Xe131_summary

# Xe129
It's two different gammas
Xe129_summary

# BiPo
there is gamma+beta and then alpha
BiPo_summary

### DIFF
--- a/fuse/plugins/micro_physics/lineage_cluster.py
+++ b/fuse/plugins/micro_physics/lineage_cluster.py
@@ -27,7 +27,7 @@ class LineageClustering(FuseBasePlugin):
     and its parent.
     """
 
-    __version__ = "0.0.26"
+    __version__ = "0.0.27"
 
     depends_on = "geant4_interactions"
 
@@ -445,11 +445,15 @@ def is_lineage_broken(
     # For gamma rays, check the distance between the parent and the particle
     if particle["type"] == "gamma":
 
-        if ((particle["creaproc"] == "phot" and particle["edproc"] == "phot") or
-            parent_lineage["lineage_type"] == 7):
+        if (particle["creaproc"] == "phot" and particle["edproc"] == "phot"):
             # We do not want to split a photo absorption into two clusters
             # The second photo absorption (that we see) could be x rays
             # So, do not split if the distance is small
+            return False
+
+        if ((parent_lineage["lineage_type"] == 7) and 
+            not (parent["creaproc"] == "RadioactiveDecayBase" 
+            and parent["edproc"] == "RadioactiveDecayBase")):
             # Same as for the previous case
             # We are already in the middle of a gamma full absorption
             # Let's not split the lineage (gamma rays, etc.)
--- a/fuse/plugins/micro_physics/lineage_cluster.py
+++ b/fuse/plugins/micro_physics/lineage_cluster.py
@@ -27,7 +27,7 @@ class LineageClustering(FuseBasePlugin):
     and its parent.
     """
 
-    __version__ = "0.0.28"
+    __version__ = "0.0.29"
 
     depends_on = "geant4_interactions"
 
--- a/fuse/plugins/micro_physics/lineage_cluster.py
+++ b/fuse/plugins/micro_physics/lineage_cluster.py
@@ -180,7 +180,7 @@ class LineageClustering(FuseBasePlugin):
                         if "[" in particle["type"]:
                             secondaries = get_all_particle_secondaries(event, particle)
                         else:
-                            secondaries = None                        
+                            secondaries = None
 
                         tmp_result = start_new_lineage(
                             particle, tmp_result, i, running_lineage_index, secondaries
@@ -197,7 +197,7 @@ class LineageClustering(FuseBasePlugin):
                     if "[" in particle["type"]:
                         secondaries = get_all_particle_secondaries(event, particle)
                     else:
-                        secondaries = None      
+                        secondaries = None
 
                     tmp_result = start_new_lineage(
                         particle, tmp_result, i, running_lineage_index, secondaries
@@ -222,11 +222,10 @@ class LineageClustering(FuseBasePlugin):
                         # New lineage!
                         running_lineage_index += 1
 
-
                         if "[" in particle["type"]:
                             secondaries = get_all_particle_secondaries(event, particle)
                         else:
-                            secondaries = None      
+                            secondaries = None
 
                         tmp_result = start_new_lineage(
                             particle, tmp_result, i, running_lineage_index, secondaries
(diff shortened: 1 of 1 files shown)


fuse/plugins/micro_physics/lineage_cluster.py
```diff
@@ -281,6 +300,24 @@ def get_parent(event_interactions, event_lineage, particle):
     return possible_parents[-1], possible_parents_lineages[-1]
 
 
+def get_all_particle_secondaries(event_interactions, particle):
+    """Returns all secondaries of the given particle.
+
+    It iterates over all interactions of the event and checks if the
+    parent id of the interaction is the same as the track id of the
+    particle, and continues down the tree until no more secondaries are
+    found or a RadioactiveDecayBase type interaction is found.
```
HenningSE, line 309: The docstring reads like the loop is supposed to check for RadioactiveDecayBase type interactions and then stop. There is no such think in the loop and I was wondering if the code needs to be updated or the docstring.

fuse/plugins/micro_physics/lineage_cluster.py
```diff
@@ -295,10 +332,20 @@ def num_there(s):
...
-def classify_lineage(particle_interaction):
+def classify_lineage(particle_interaction, secondaries=None):
     """Function to classify a new lineage based on the particle and its parent
     information."""
 
+    # Check if we passed secondaries
+    if secondaries is not None:
+        for secondary in secondaries:
+            # there is a seconday that has type gamma and edproc phot,
+            # we classify this lineage as beta because it is most likely a compton scattering
```
HenningSE, line 343: This comment is also maybe outdated as the code does not check for gamma and phot edproc?

fuse/plugins/micro_physics/lineage_cluster.py
```diff
@@ -362,7 +409,6 @@ def classify_lineage(particle_interaction):
         return NEST_NONE
 
 
-@njit()
```
HenningSE, line 365: Did you check how the speed of the code changes when removing numba? (Just want to make sure we do not slow done too much)

fuse/plugins/micro_physics/lineage_cluster.py
```diff
@@ -372,18 +418,46 @@ def is_lineage_broken(
...
         return True
 
     # For gamma rays, check the distance between the parent and the particle
     if particle["type"] == "gamma":
 
+        if particle["creaproc"] == "phot" and particle["edproc"] == "phot":
+            # We do not want to split a photo absorption into two clusters
+            # The second photo absorption (that we see) could be x rays
+            # So, do not split if the distance is small
+            return False
```
HenningSE, line 453: Here you do not evaluate the distance between the energy deposits? Please update the comment or code :-)

fuse/plugins/micro_physics/lineage_cluster.py
```diff
@@ -393,14 +467,13 @@ def is_lineage_broken(
...
-        if (particle["creaproc"] == "phot") and (particle["edproc"] == "phot"):
-            # we do not want to split a photo absorption into two clusters
-            # the second photo absorbtion (that we see) could be x rays
-            # so, do not split if the distance is small
-            return False
+        if particle["creaproc"] == "eBrem":
+            # we do not want to split a bremsstrahlung into two clusters
+            # if the distance is really small, it is most likely the same interaction
+            if distance < 0.1:  # cm
+                return False
```
HenningSE, line 476: I think we should put this into a URLConfig to keep track of it.

fuse/plugins/micro_physics/lineage_cluster.py
cfuselli: Thanks Henning, I had it like that at the beginning but then realised I need only first generation secondaries. Updating docstring now :)

fuse/plugins/micro_physics/lineage_cluster.py
cfuselli: Yes, also from previous version of the code. Thanks!

fuse/plugins/micro_physics/lineage_cluster.py
cfuselli: Not checked yet.. but on small runs I saw no notable changes.

fuse/plugins/micro_physics/lineage_cluster.py
cfuselli: Added :)

cfuselli:
Btw I am still discussing with Micha about this PR and we are still not sure how we want to treat those cases.

cfuselli:
With the latest commit we do a bit of "dietrofront" on the treatment of RDB decays. 
I will update the description of the PR asap. 

New treatment involves: 
- merge together all electrons coming from excited stated RDB decays, they are conversion electrons, and we give them a beta cluster. 
- keep doing the "if gamma break lineage". when we have both conversion electrons and gammas, we want to let the gammas travel and get a compton (beta yield) or photabs (gamma yield) cluster. 
- we remove the need to classify based on secondaries and all the cases can be handled with just a few lines, specifying to not break if there is a "[" in the parent and to assing an 8 if there is an "[" in the type.

PR #257, opened 2024-09-09
Take yield and width parameters from config file

HenningSE:

This PR will let fuse take the yield parameters from a simulation config file.


cfuselli (approved):
Looks good!

PR #260, opened 2024-09-11
pulse_id type casting inconsistency

SaelMo:
While checking out why I am getting `numpy.exceptions.DTypePromotionError: field titles of field 'pulse_id' mismatch` I found this error with Henning and after changing the casting to be consistent it seems to have vanished.


HenningSE (approved):
Thanks for fixing this!

PR #261, opened 2024-09-12
Make sure CDF goes to 1 smoothly

HenningSE:

This PR makes sure that the CDF for the SPE waveform is properly normalized.


ramirezdiego (approved):
Great catch, thanks a lot.

PR #262, opened 2024-09-12
Option to replace maps by constant dummy map

HenningSE:

This PR adds a URLConfig protocol, creating a dummy map returning constant values. The DummyMap code is taken from [WFSim](https://github.com/XENONnT/WFSim/blob/b75494fbd594c329eee54535b0bb9754bc7458d8/wfsim/load_resource.py#L437-L457).


ramirezdiego (approved):
Thanks, @HenningSE! Merging now.

PR #269, opened 2024-10-11
Corrections override

HenningSE:

This PR will replace the `CorrectedAreas` plugin from straxen with a modified one for MC data. It will set `cs1` and `cs2` to the values of `cs1_wo_timecorr` and `cs2_wo_timecorr` as we do not simulate these effects.


HenningSE:
Besides the new plugin, this PR removes unused config arguments from the context. This should reduce the number of warnings during the simulation.

GiovanniVolta:
Hey @HenningSE, thank you very much. It looks good to me. I would just ask you to add in the plugin docstring the motivation for this

ramirezdiego (approved):
Thank you, Henning!

PR #270, opened 2024-10-11
Remove microphysics_context

HenningSE:

This PR will remove the `microphysics_context`. All simulations can be run with the `full_chain_context` and we do not need to keep this code.


ramirezdiego:
@HenningSE Ready for review?

PR #273, opened 2024-10-17
Little bugfix for CSV generators

ramirezdiego:
The `trackid` in Geant4 starts always at `1`, so that `parentid` can be assigned to `0` for the primary particles. If we set both to `1`, the `get_parent` function [here](https://github.com/XENONnT/fuse/blob/36ac25c5648bfd5702c8a22101f8e338547b0a10/fuse/plugins/micro_physics/lineage_cluster.py#L334C5-L334C15) gets confused and gives us a wrong NEST id for the evaluated interaction (hence we eventually reconstuct wrong energies along the simulation chain).

See screenshot before and after this change, in a context where `clustering_method="lineage"` and where gammas are simulated in our CSV input. Only the second case gives us the right `lineage_type` expected for these intereactions (NEST_id = 8; beta model):
image

In principle lineage clustering plays no role if we simulate single scatters, as is the case for the generators in this notebook, but this quick fix prevents confusing people when it becomes our default clustering method.


HenningSE (approved):
Thanks for the PR @ramirezdiego! As you mentioned, the lineage clustering can be a bit tricky with CSV input so one should always check if the algorithm works with the input instructions. For most cases, DBSCAN should give us the same results but the classification of clusters might differ a bit.

PR #282, opened 2024-12-28
Be compatible with straxen >= 3

dachengx:
But before this, there should be a final stage release of fuse with strax<2, straxen<3.


fuse/plugins/detector_physics/csv_input.py
```diff
@@ -69,6 +69,18 @@ class ChunkCsvInput(FuseBasePlugin):
         help="n_interactions_per_chunk",
     )
 
+    first_chunk_left = straxen.URLConfig(
```
dachengx, line 72: This is just putting these parameters in lineage. In principle, they are not related to this PR.

PR #287, opened 2025-01-26
Only simulate a part of the PI train

dachengx:

`photoionization_time_cutoff` is the cutoff when modeling `photoionization_modifier`. `photoionization_modifier `will be smaller when cutoff is smaller because analysts only look at a part of the distribution not the whole one.

`photoionization_time_cutoff_mc` is the cut off choosen to simulate only a part of the photoionization distribution to make simulation faster. By setting `photoionization_time_cutoff_modeling` smaller than `photoionization_time_cutoff_mc`, you will simulate "less PI". The PI strength or time constant will not be changed.


HenningSE (approved):
Thanks for the nice PR, looks good to me!

PR #289, opened 2025-02-06
updating the fixed time separation

HenningSE:

The PR fixes a problem with the time calculation in the fixed spacing option. Up to now we use `np.arange` with a fixed step size to produce the event times. This caused a shape mismatch as `num_interactions+1` times are generated. With this PR, `np.linspace` will be used to generate a matching number of times.


cfuselli:
Hey @HenningSE , 
I gave a quick try with both functions, and to me it seems like they both produce `num_interactions` points. 
See here: 



```
Testing functions:
num_interactions: 10
event_rate: 0.1
--------------------------------------------------
Using np.arange (current)
10 interaction times: [ 0. 10. 20. 30. 40. 50. 60. 70. 80. 90.]
Using np.linspace (proposed)
10 interaction times: [  0.          11.11111111  22.22222222  33.33333333  44.44444444
  55.55555556  66.66666667  77.77777778  88.88888889 100.        ]
```

Or am I missing something? 



[ output produced with: ]
```
import numpy as np 
num_interactions = 10
event_rate = 0.1
a = np.arange(
    start=0, stop=num_interactions / event_rate, step=1 / event_rate
    )
b = np.linspace(
    start=0, stop=(num_interactions) / event_rate, num=num_interactions
)

print("Testing functions:")
print(f"num_interactions: {num_interactions}")
print(f"event_rate: {event_rate}")
print("-"*50)
print("Using np.arange (current)")
print(f"{len(a)} interaction times: {a}")
print("Using np.linspace (proposed)")
print(f"{len(b)} interaction times: {b}")
```

HenningSE:
Hi @cfuselli you are right. I can not produce the error right now, maybe it was just a strange combination of numbers when it showed up. I will have a closer look next time I run some simulations.

cfuselli:
Triggered by Henning that encountered the issue again, we realised that np.arange can have floating point errors that can lead to unexpected behaviour. Quoting from [numpy docs](https://numpy.org/doc/stable/reference/generated/numpy.arange.html) .

> The length of the output might not be numerically stable.
> Another stability issue is due to the internal implementation of [numpy.arange](https://numpy.org/doc/stable/reference/generated/numpy.arange.html#numpy.arange). The actual step value used to populate the array is dtype(start + step) - dtype(start) and not step. Precision loss can occur here, due to casting or due to using floating points when start is much larger than step. This can lead to unexpected behaviour. For example:
> In such cases, the use of numpy.linspace should be preferred.

So we should indeed prefer the use of np.linspace here.

cfuselli (approved):
Approved.

PR #296, opened 2025-03-04
Removed electron drift electrons starting at z > -1 cm

sebvetter:
The default field distortion map only extends up to z = -1 cm, afterwards it extrapolates. Electrons at the very top of the TPC do not drift and need to field distortion. Therefore I removed the application of the FDC for z > -1 cm.

Make a selection of events with z < -1cm and apply the field distortion on the selected events (instead of all).

electron_drift_use_fdc_for_top
Before the fix: The position in the instructions and the position at the interface are not the same because of electron drift.

electron_drift_no_fdc_for_top
After the fix: The instructed position and the interface position are the same (only electrons starting at z = -0.1cm shown).


sebvetter:
Not yet, the boundary is currently hardcoded. The better way, in my eyes, is to set the fill value for out of bounds points of the interpolator to 0. I will try and get to testing this in the near future

cfuselli:
Closed in favour of #304

Issue #295, opened 2025-02-26
Sorting Issue: pulse_id and Time Become Misaligned After Sorting


zhut19:
I tried to simulate a scenario where digitizers are triggered externally by forcing pulse windows across channels to start simultaneously. However, I encountered misalignments in the digitization process. Specifically:

In [this section](https://github.com/XENONnT/fuse/blob/7a651c3ea10af819cf3183fcc9034c212e073044/fuse/plugins/pmt_and_daq/pmt_response_and_daq.py#L565), `propagated_photons` are no longer ordered in time afterward, as the default quick sort won't preserve the original order when the compare is the same.
In [this section](https://github.com/XENONnT/fuse/blob/7a651c3ea10af819cf3183fcc9034c212e073044/fuse/plugins/pmt_and_daq/pmt_response_and_daq.py#L232), that `pulse_groups` is not sorted by `pulse_id`, leading to a mismatch between elements in `pulse_groups` and `photons`.

PR #297, opened 2025-03-10
Daq photon sorting

zhut19:
## Fixing #295 

Use `strax.stable_sort` instead of `np.sort`, and add sorting to `pulse_groups` as well.


HenningSE (approved):
Great! Thanks @zhut19 for the PR!

PR #298, opened 2025-03-14
Fuse compatibility with straxen (support both straxen v2 and v3)

cfuselli:
After https://github.com/XENONnT/fuse/pull/282 we were not compatible with older strax and straxen. 
Now we address changes needed for being compatible with straxen v2.* and v3.*

-------

- Check if xnt_common_opts exists or just common_opts, choose accordingly
- Only do purge_configs if supported
- Go back to support older strax(en). Now strax = ">=1.6.0" and straxen = ">=2.2.3"
- Additionally, we include new auto tests to make sure we stay compatible with strax = "1.6.0" and straxen = "2.2.3"
- We add also the stable sort function as in https://github.com/AxFoundation/strax/pull/918


cfuselli:
Fine to close this one in favour of #299 that contains the same commits.

HenningSE:
Thanks! Looks good to me. I will test it tomorrow and if it works fine we can merge.

HenningSE (approved):
Thanks @cfuselli!

PR #299, opened 2025-03-26
Move context to fuse

cfuselli:
Proposal to move context definition to fuse. 

For details on the proposal, see [wiki note](https://xe1t-wiki.lngs.infn.it/doku.php?id=cfuselli:restructure_fuse_contexts).

### DIFF
--- a/fuse/context.py
+++ b/fuse/context.py
@@ -18,7 +18,7 @@ if hasattr(straxen.contexts, "xnt_common_opts"):
     common_opts = straxen.contexts.xnt_common_opts
     common_config = straxen.contexts.xnt_common_config
 else:
-    # This is for straxen >=3, variable names changed    
+    # This is for straxen >=3, variable names changed
     common_opts = straxen.contexts.common_opts
     common_config = common_config
 
@@ -111,9 +111,7 @@ def microphysics_context(
 
     st = strax.Context(storage=strax.DataDirectory(output_folder), **common_opts)
 
-    st.config.update(
-        dict(detector="XENONnT", check_raw_record_overlaps=True, **common_config)
-    )
+    st.config.update(dict(detector="XENONnT", check_raw_record_overlaps=True, **common_config))
 
     # Register microphysics plugins
     for plugin in microphysics_plugins_dbscan_clustering:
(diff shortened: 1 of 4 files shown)


HenningSE:
Ohh what? Copilot is reviewing PRs now?

fuse/context.py
```diff
@@ -231,106 +241,75 @@ def full_chain_context(
...
+    It takes the general full_chain_context and sets the proper
+    corrections and configuration files for XENONnT.
+    """
+
+    import fuse
+
+    # Get the simulation config file
+    if "/" in simulation_config:
+        # Local path to the simulation config file
+        simulation_config_file = simulation_config
```
HenningSE, line 273: Maybe we are a bit more flexible if we check with something like: `os.path.isfile` if the config file is a local one or not.

fuse/context.py
```diff
@@ -231,106 +241,75 @@ def full_chain_context(
...
+
+    # Get clustering method
+    # if it is specified as an argument, use that
+    # if it is not specified, try to get it from the config file
+    # if it is not in the config file, use dbscan
+    if clustering_method is None:
+        try:
+            clustering_method = fuse.from_config(simulation_config_file, "clustering_method")
+        except ValueError:
+            clustering_method = "dbscan"
```
HenningSE, line 295: I think it would be nice if we print a little info to the log which clustering method was used. Just to make sure the user is aware what was picked in the end.

fuse/context.py
```diff
@@ -231,106 +241,75 @@ def full_chain_context(
...
+        output_folder=output_folder,
+        corrections_version=corrections_version,
+        simulation_config_file=simulation_config_file,
+        corrections_run_id=corrections_run_id,
+        clustering_method=clustering_method,
+        **kwargs,
+    )
+    st.set_config(old_xedocs_versions_patch(corrections_version))
+
+    fdc_ext = fdc_map_mc.split(fdc_map_mc.split(".")[0] + ".")[-1]
```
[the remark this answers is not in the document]
cfuselli: wrong suggestion.. sorry copilot

HenningSE (approved):
For me it is working fine! Thanks @cfuselli for the fantastic refactoring of the contexts!

PR #300, opened 2025-03-31
Be compatible still with strax 1 + straxen 2

dachengx:
https://github.com/XENONnT/fuse/pull/282 is too strict in the strax(en) version. This PR checks the attribute of `straxen.contexts` and assign configurations of context accordingly.

### DIFF
--- a/fuse/context.py
+++ b/fuse/context.py
@@ -93,16 +93,30 @@ processing_plugins = [
 ]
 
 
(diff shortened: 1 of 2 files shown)


cfuselli:
Hi @dachengx thanks, I think this was being addressed in the same way here https://github.com/XENONnT/fuse/pull/298

PR #301, opened 2025-03-31
Cleanup unused configs

cfuselli:
some configs are not used in the plugins


HenningSE:
Hmmm... Interesting. Maybe some copy-paste thing that was never spotted? I don't see a reason to keep it. Thanks for spotting this @cfuselli. We can include it in the next release with the other open PRs.

HenningSE (approved):

PR #302, opened 2025-04-03
Patch cmake version in nestpy

cfuselli:
All fuse tests fail now because of a problem with cmake . 
Probably due to a change in github actions cmake version (see [here](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#command:cmake_minimum_required)).

With this PR we patch the pybind version to v2.13.0 and change the cmake_minimum_required(VERSION 2.8.12...3.30) in the nestpy cmake list. 

See also https://github.com/NESTCollaboration/nestpy/issues/113

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -48,6 +48,10 @@ jobs:
           cd nestpy
           git submodule update --init --recursive
           git checkout fb3804e
+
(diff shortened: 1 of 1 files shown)


HenningSE (approved):
Thanks for making the tests work again!

PR #304, opened 2025-04-17
Clip FDC input positions

cfuselli:
This PR is an alternative solution to #296.

As suggested by @sebvetter , a better solution than hardcoding the value of Z for the FDC map boundaries as in #296 , here we clip the positions that we pass to the map to the boundaries of the validity of the map (that we get from the map coordinate system). 

The results seem to give the required behaviour, here comparing clipped vs unclipped DeltaR as a function of Z, for a given R. 

output_r0
output_r20
output_r60

### DIFF
--- a/fuse/plugins/detector_physics/electron_drift.py
+++ b/fuse/plugins/detector_physics/electron_drift.py
@@ -342,7 +342,8 @@ class ElectronDrift(FuseBasePlugin):
         )
         if n_clipped_r > 0 or n_clipped_z > 0:
             self.log.warning(
-                f"Field distortion map is clipped {n_clipped_r} times in r and {n_clipped_z} times in z"
+                f"Field distortion map is clipped {n_clipped_r} \
+                    times in r and {n_clipped_z} times in z"
             )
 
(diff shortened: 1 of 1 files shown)


cfuselli:
Thanks Henning!
Added the logging with: 
` f"Field distortion map is clipped {n_clipped_r} times in r and {n_clipped_z} times in z"
`

sebvetter:
The drift simulation works, however there is still some minor drift for events at the very top of the TPC:
image

This is not an issue for everything except posrec simulations I think. I also think this method is more realistic than my simply hardcoded 0 drift at this height. Just to note, this is still not the setting we want to use for psorec simulations. But Carlo was so kind to point me to [this line](https://github.com/XENONnT/fuse/blob/0705e475edd7b3e5cf963fc5eef663db59144fcb/fuse/plugins/detector_physics/electron_drift.py#L239): we can simply set the map name to None and avoid simulating drift for the posrec simulations. For every other use-case (that I could think of), this PR should be he appropriate solution.

cfuselli:
Thanks @HenningSE and @sebvetter for your comments. 
Seems all good so I will merge :)

PR #308, opened 2025-05-22
MC override in context config

cfuselli:
Allow flexible overwrite of straxen configs from fuse config file. 

I will remove the fdc_map_mc from the new config format, and leave the old way of treating the fdc map in the fuse context.
In this way the new fuse version would be compatible with both newer and older config files, but old fuse versions not compatible with new config file (potentially dangerous as it would ignore the overwrite).

Supports formats like: 

```
"mc_overrides": {
  "s1_xyz_map": ["XnT_S1_xyz_123.json", "itp_map://resource://MAP_NAME?fmt=json"]
  "fdc_map": "itp_map://resource://XnT_S1_xyz_123.json?fmt=json"]
}

```
in the config file, as in https://github.com/XENONnT/private_nt_aux_files/pull/424.

### DIFF
--- a/fuse/context.py
+++ b/fuse/context.py
@@ -324,7 +324,7 @@ def xenonnt_fuse_full_chain_simulation(
     st.set_config({"fdc_map": fdc_conf})
 
     # Overwrite some resource files with the ones from the simulation config
-    apply_mc_overrides(st, corrections_version)
+    apply_mc_overrides(st, simulation_config_file)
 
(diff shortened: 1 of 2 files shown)


HenningSE (approved):
Looks good to me, Thanks @cfuselli !

ramirezdiego (approved):
+1!

cfuselli:
Thank youuu!

PR #310, opened 2025-05-26
Fix context to be compatible with new config formats

cfuselli:
With #308 I introduced a new context to be compatible both with old and new configs.
It should have supported, for example, setting the fdc_map both with fdc_map_mc or mc_overrides.
But the logic was wrong so the fdc_map_mc was still required, making the new context not usable. 
Now this should be fixed in this PR, so we can still use both formats.


ramirezdiego:
Nice! :) New release?

PR #316, opened 2025-06-05
Migdal yield

lorenzomag:

This code implements the Migdal effect when computing the amount of quanta produced by a nuclear recoil (NR).
Currently, no charge yield suppression is considered.

The code utilises a pickled class containing interpolators for the probability tables provided by [Peter Cox _et al._] (https://petercox.github.io/Migdal/) to compute the probability of producing a Migdal ionisation as a result of a NR.

It also requires additional fields to the simulation config file; these include the mass of the Xe atom, the binding energies for its orbitals, a boolean to force a Migdal event for every NR, which orbitals to consider for the simulation.

These additional files are currently available in this [working repository](https://github.com/XENONnT/migdal-yield.git).

Using simple MC techniques, the code determines, for each NR event, if there was a Migdal event, which orbital the ionisation electron is coming from, and its ionisation energy.
This energy and the energy of relaxation X-rays and Auger electrons (which sum up to the binding energy of the electron) are then converted to quanta using NEST β-yield model.

The quanta produced by this ER interaction are then simply added to the ones arising from the corresponding NR. This algorithm will have to be improved to account for charge yield suppression.

### DIFF
--- a/tests/test_MicroPhysics.py
+++ b/tests/test_MicroPhysics.py
@@ -114,5 +114,6 @@ class TestMicroPhysicsAlternativePlugins(TestMicroPhysicsBase):
         )
         self.test_context.make(self.run_number, "quanta")
 
+
 if __name__ == "__main__":
     unittest.main()


fuse/plugins/micro_physics/yields.py
```diff
@@ -24,7 +26,7 @@ class NestYields(FuseBasePlugin):
     provides = "quanta"
     data_kind = "interactions_in_roi"
 
-    dtype = quanta_fields + strax.time_fields
+    dtype = strax.time_fields + cluster_id_fields + quanta_fields
```
cfuselli, line 29: Hey! Here I think we should leave it with only the quanta fields as it was. Youd don't really need to do it, and if you need those fields in some later plugin you can just load together the quanta and the interactions_in_roi.

fuse/plugins/micro_physics/yields.py
```diff
@@ -131,6 +133,9 @@ def compute(self, interactions_in_roi):
         result = np.zeros(len(interactions_in_roi), dtype=self.dtype)
         result["time"] = interactions_in_roi["time"]
         result["endtime"] = interactions_in_roi["endtime"]
+        result["cluster_id"] = interactions_in_roi["cluster_id"]
```
cfuselli, line 136: so this one also goes. I think the nest yields plugin can stay untouched.

fuse/plugins/micro_physics/yields.py
```diff
@@ -448,3 +453,327 @@ def get_NR_quanta(self, energy, field, par_dict):
...
+
+
+@export
+class MigdalYields(NestYields):
+    __version__ = "0.0.1-alpha"
+    child_plugin = True
+
+    provides = ("quanta", "migdal_truth")
+
+    common_dtypes = strax.time_fields + cluster_id_fields
```
cfuselli, line 465: even here cluster id I think it can go!

fuse/plugins/micro_physics/yields.py
```diff
@@ -448,3 +453,327 @@ def get_NR_quanta(self, energy, field, par_dict):
...
+    dtype = {
+        "quanta": quanta_dtypes,
+        "migdal_truth": truth_dtype,
+    }
+
+    data_kind = {data_type: "interactions_in_roi" for data_type in provides}
+
+    xenon_mass = straxen.URLConfig(
+        default="simple_load://simulation_config://"
+        "SIMULATION_CONFIG_FILE.json?"
```
cfuselli, line 500: good. we will have to add all these numbers to the sim config file. if these are standard values you can also just set the value as default, without having to use the config file. not sure..

fuse/plugins/micro_physics/yields.py
```diff
@@ -448,3 +453,327 @@ def get_NR_quanta(self, energy, field, par_dict):
...
+        default="simple_load://resource://"
+        "Migdal_probability_distribution_interpolators.pkl?"
+        "&fmt=pkl",
+        help="Interpolators for the Migdal distribution functions. "
+        "Migdal class instance from https://github.com/petercox/Migdal/blob/v1.0.0/Migdal.py .\n",
+    )
+
+    force_migdal = straxen.URLConfig(
+        default="simple_load://simulation_config://"
+        "SIMULATION_CONFIG_FILE.json?"
```
cfuselli, line 533: this one also I think probably no need to go in the config file? can default to true or false, and when the user wants to change it it can just change in the st.set_config

cfuselli (changes requested):
Hey Lorenzo thanks a lot for this plugin! Looks good to me, I requested a few changes but overall looks all good. Maybe one more thing we should add a test for this plugin?

lorenzomag:
Ok, my fix didn't work. I might have to add typing info to NestYields, unless someone has power over pre-commit 🙃

PR #321, opened 2025-07-03
Skip aft correction for fuse? 

cfuselli:
At lease if we do not simulate PI, do not correct AFT for PI effect.

Not sure if this is the best way to do it.


cfuselli:
This is not a good way to solve this

PR #326, opened 2025-07-08
Allow None `corrections_run_id`

dachengx:

This PR confines the code for writing `run_id` into lineage within a single function, `write_run_id_to_config`.

The PR allows `corrections_run_id` to be None, by setting `run_without_proper_run_id=True`.

### DIFF
--- a/fuse/context.py
+++ b/fuse/context.py
@@ -144,6 +144,7 @@ def full_chain_context(
         "drift_time_gate": "electron_drift_time_gate",
     },
     run_without_proper_corrections=False,
+    run_without_proper_run_id=False,
     clustering_method="dbscan",
     extra_plugins=[],
 ):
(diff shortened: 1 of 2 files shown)


dachengx:
> Hey Dacheng looks all good to me here. The only thing I am not sure is the name of the `run_without_proper_run_id`. This is what you want to use for saltax right? So for a saltax user, setting `run_without_proper_run_id` would sound a bit strange I guess - as it does almost the opposite thing?
> 
> Maybe I misundertand the use, but if I understand correctly the effect of this is kind of a: "use the run id as corrections run id"?

It is following the naming of `run_without_proper_corrections`. And the `run_without_proper_run_id=False` means that we will use the assigned default `corrections_run_id`. And to avoid change on the result outside saltax, `run_without_proper_run_id=True` is not suggested.

cfuselli (approved):
Looks good thanks Dacheng. Let's make saltax's use easier :)

PR #328, opened 2025-07-16
Do not decorate `run_id=` by the default of plugin configs

dachengx:
https://github.com/XENONnT/fuse/pull/326 made a bug

`run_id=` should be revised only based on `st.config`, but not the default of plugin configs

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
         additional_dependencies: [types-PyYAML, types-tqdm]
 
 -   repo: https://github.com/pycqa/doc8
-    rev: v2.0.0
+    rev: v1.1.2
     hooks:
     -   id: doc8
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -29,7 +29,7 @@ repos:
         additional_dependencies: [types-PyYAML, types-tqdm]
 
 -   repo: https://github.com/pycqa/doc8
-    rev: v2.0.0
+    rev: v1.1.2
```
cfuselli, line 32: was this intentional?

cfuselli (approved):
Bug seems fixed thanks Dacheng. 
Maybe just remove the change in precommit if not intentional. 
Thanks!

.pre-commit-config.yaml
dachengx: it is intentional

doc8 v2.0.0 needs python >= 3.10

PR #330, opened 2025-07-22
Bump to v1.5.5

dachengx:


cfuselli:
Actually, @dachengx we usually don't put usernames in the HISTORY.md (not sure why). 
If you don't mind, can you remove them so it's following the style of previous releases?

PR #333, opened 2025-07-25
Add preferred interpolation method for s2 patter map

cfuselli:
We give the option to choose the interoplation method of the s2 pattern map by setting a value in the key of the map called `preferred_interpolation_method`. 
If the key is not present, still the default will be used.


cfuselli:
Closing in favour of #334

PR #334, opened 2025-07-25
Add s2_pattern_map_interpolation_method

cfuselli:
We want to be able to define the interpolation method of the s2 pattern map. 
We leave as default the `WeightedNearestNeighbors`. 
We will change it accordingly in the simulation config files using the mc_overrides key.


cfuselli:
Tested and gives: 
`DEBUG:fuse.context_utils:Interpolating pattern map 'XENONnT S2 LCE per-PMT map' with method 'Linear1Din2D'
`with SR2 configs and 
`DEBUG:fuse.context_utils:Interpolating pattern map 'XENONnT S2 LCE per-PMT map' with method 'WeightedNearestNeighbors'
`with SR1 configs.

HenningSE (approved):
The new configs were already updated accordingly, so all good from my side!

PR #336, opened 2025-08-16
Use stable `sort` and `argsort` to avoid indeterministic result

dachengx:
This PR was initiated when I found that the `propagated_s1_photons` and `propagated_s2_photons` are not deterministic, i.e., the result will be a little bit different each time I submit a job.

`np.sort` and `np.argsort` in fuse are only partially changed to `stable_sort` and `stable_argsort`. This PR applies stable sorting to all sorting in fuse.

Depends on https://github.com/AxFoundation/strax/pull/918 and https://github.com/XENONnT/fuse/issues/298

### DIFF
--- a/fuse/common.py
+++ b/fuse/common.py
@@ -28,8 +28,8 @@ kind_colors = dict(
 
 @numba.njit()
 def dynamic_chunking(data, scale, n_min):
-    idx_sort = np.argsort(data)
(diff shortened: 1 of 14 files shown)


cfuselli (approved):
Thanks Dacheng this looks all good to me.

PR #340, opened 2025-09-08
Fix bug caused by event id time overlapping in `interaction_lineage `

dachengx:

Different `eventid` can overlap in time of `geant4_interactions`. `geant4_interactions` is sorted in time, but the returned values of `build_lineages` are sorted in `eventid`.

Sort `geant4_interactions` by `eventid` and time, run `build_lineages`, and finally recover the order of `interaction_lineage` according `geant4_interactions`.


cfuselli (approved):
Good catch thanks @dachengx . 
I tested and seems to work well for me. 
Also, we do sort again inside build_lineages so the logic of the function is not affected, but the bug reported is fixed.

PR #346, opened 2025-09-16
Optimize clustering peformance

WenzDaniel:

Optimized `cluster_and_classify` to only loop once over all clusters. I compared the output of the new and old function using 25k neutrons using np.isclose:

```python
for field in result.dtype.names:
    print(field, np.all(np.isclose(result[field], result2[field])))
```

I profiled the performance increase using 10k neutrons distributed over 15 chunks. The call of time of `cluster_and_classify` reduced from 79 s to 0.7 s. The performance increase becomes even more significant when moving to bigger chunks.

### DIFF
--- a/fuse/plugins/micro_physics/merge_cluster.py
+++ b/fuse/plugins/micro_physics/merge_cluster.py
@@ -1,6 +1,7 @@
 import strax
 import straxen
 import numpy as np
+import numba
 
 from ...dtypes import (
     primary_positions_fields,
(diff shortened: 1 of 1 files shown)


cfuselli:
Looks good thanks @WenzDaniel . 
If you can apply the sable argsort as Dacheng mentioned I think we can merge it. 
Thanks!!

dachengx (approved):

PR #348, opened 2025-09-18
Optimized merge lineage performance

WenzDaniel:

Same as #346, but for the lineage clustering. I tested the new code using the following settings:

```python
st = fuse.context.xenonnt_fuse_full_chain_simulation(
    output_folder="/project2/lgrandi/wenz/strax_data/test_data/",
    corrections_version="global_v18",
    simulation_config="sr2_dev",
    clustering_method="lineage"
)



st.set_config(
    {
        "path": "/scratch/midway3/myoshida1/sr2_nr_bg/MC/geant4_outputs/",
        "file_name": "AmBe_BottomCW6d9m_wo_gamma_0.root",
        "entry_start": 0,
        "entry_stop": 10_000,
        "fixed_event_spacing": True,
        "source_rate": 0.5,
        "separation_scale": 10**9,
        "n_interactions_per_chunk": 500_000,
        "nr_only": False,
    }
)
```

The new function reduced computation time of `merge_lineages` from 40.6 s to 0.02 s reducing the wall computation time of the `microphyscics_summary` by a factor 1.4. I tested that both codes return the same clusters using

```python
for name in clustered_interactions_vanilla.dtype.names:
    print(name, np.all(np.isclose(clustered_interactions_vanilla[name], clustered_interactions_optimized[name])))
```

although I had to first sort both arrays by `cluster_id` given that the new code requires the data to be sorted by cluster_id.

### DIFF
--- a/fuse/plugins/micro_physics/merge_cluster.py
+++ b/fuse/plugins/micro_physics/merge_cluster.py
@@ -1,6 +1,7 @@
 import strax
 import straxen
 import numpy as np
+import numba
 
 from ...dtypes import (
     primary_positions_fields,
@@ -9,6 +10,7 @@ from ...dtypes import (
     cluster_misc_fields,
 )
 from ...plugin import FuseBasePlugin
+from ...common import stable_sort
 
 export, __all__ = strax.exporter()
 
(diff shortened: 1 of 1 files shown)


cfuselli:
Same, looks very good thanks @WenzDaniel !

PR #361, opened 2025-10-22
Modified load_root_file for supporting multiple primary positions in G4

APaloma710:
Added code to check whether root files have more than one primary position per G4 event. If that's the case, only the first (x_pri, y_pri, z_pri) set will be used.


mhliu0001:
Hi @APaloma710 , I am wondering what the status is here, since I also need this feature for the AmBe simulation to work. If you do not object, I will fix this PR to make it work.

PR #371, opened 2026-01-29
Also provide energy truth for S1 and S2

WenzDaniel:
Provide additional truth information about the S1 and S2 energy instead of just the mean of the two.


cfuselli (approved):
Looks good to me

PR #384, opened 2026-05-15
Pr/lineage clustering

DonNabla:
## Summary

  Three-step performance refactor of `LineageClustering.build_lineage_for_event`:
  eliminate redundant `np.where` rescans, pre-compute the string-field
  comparisons as integer codes, and lift the per-particle inner loop into
  a single `@numba.njit` kernel.

  The `LineageClustering` plugin's per-event wall drops by roughly 5×
  on realistic γ-background simulations. Bit-identical to the unmodified
  plugin (44 M `interaction_lineage` rows checked across a multi-source
  MC sweep, plus a frozen baseline fixture).

  > **Depends on #383** (`fix/get-parent-same-time-tag`).
  > This PR is branched off `fix/get-parent-same-time-tag` so its numba
  > kernel inlines the spatial tie-break logic introduced there. Once
  > #383 is merged, this branch rebases trivially onto `main`.

  ## What changes

  Three logically independent commits, each individually bit-identical
  to its predecessor:

  1. **`perf(lineage_cluster): eliminate O(N) rescans inside build_lineage_for_event`**
     - `get_parent` and `get_last_particle_interaction` now take a
       precomputed `trackid_lookup` (built once at the top of
       `build_lineage_for_event`) and fetch the parent's row indices
       in O(1) instead of re-running `np.where(trackid == parent_id)`
       for every interaction.
       set tracked inline, dropping a per-particle `np.all` scan.
       
  2. **`perf(lineage_cluster): pre-tag string fields with int enums`**
     - The per-particle classification rules previously did string
       compares on `type`, `parenttype`, `creaproc`, `edproc` for every
       row. This commit computes an `int8` code per row once at the
       top of `compute()` (unknown strings encode as `-1` so they
       never match a named constant), then runs `_is_broken_coded`
       and `_classify_coded` against the int codes.
     - Also lifts `precompute_ion_AZ` out of the hot path (regex is
       applied once per unique `type` string rather than per row),
       and inlines `start_new_lineage` / `continue_lineage` as direct
       field-view writes on the result buffer.

  3. **`perf(lineage_cluster): numba-njit the build_lineage_for_event inner loop`**
     - The Python per-particle loop is replaced by a single
       `@numba.njit` kernel `_build_lineage_for_event_kernel`. The
       wrapper builds a CSR-style trackid lookup (`trackid_offsets`,
       `trackid_indices`, `trackid_pos`, `parent_pos`) that replaces
       the Python dicts; the kernel walks the per-particle loop
       entirely in compiled code, calling `_classify_njit` and
       `_is_broken_njit` (int-only twins of the previous Python
       helpers) for each interaction.
     - `main_cluster_type` stays on the numpy path because numba
       handles structured-string outputs poorly and it is not in the
       inner-loop hot path.
     - The kernel inlines the parent-finding logic. Following the
       spatial tie-break from #383, the inlined logic does two
       passes: (1) find `t_max = max(t[idx] for idx in slice if
       t[idx] <= t_i)`, (2) among entries with `t == t_max`, pick the
       one closest to the daughter's `(x, y, z)`. Squared distance is
       used to avoid a sqrt — `argmin` is the same. A nearest-`|t -
       t_i|` fallback handles the case where no parent step has
       `t <= t_i`.
  
  ## Bit-identity verification
  
  - New unit equivalence test
    `tests/test_LineageClustering_equivalence.py` asserts
    `np.array_equal` field-by-field on `interaction_lineage` against a
    frozen baseline (`tests/data/lineage_baseline.npz`, 26 306 rows on
    the standard test fixture). Test passes in ~6 s.
  - Out-of-tree: `interaction_lineage` output was also compared field-
    by-field against vanilla 1.6.3 + #383 on `K40`, `Co60`, `Th232`,
    `Ra226`, and `Pb212` simulated runs — 44 M total rows, all
    identical (and identical between each pair of consecutive perf
    commits, since each step preserves output bit-by-bit).

  ## Performance
  
  Per-source `interaction_lineage` target wall, before → after, on
  realistic Geant4 simulations:

  | source | vanilla (s) | this PR (s) | speedup |
  |---|---:|---:|---:|
  | K40   |  149 |  30 | 5.0× |
  | Co60  | 1920 | 381 | 5.0× |
  | Th232 |  104 |  27 | 3.9× |
  | Ra226 |  140 |  28 | 5.0× |
  | Pb212 | 2271 | 932 | 2.4× |

  (Pb212 has the smallest relative gain because its `interaction_lineage`
  target is dominated by Geant4 ROOT reading + upstream clustering rather
  than the per-particle Python loop this PR optimises.)
  
  ## Notes
  
  - No upstream code outside `fuse/plugins/micro_physics/lineage_cluster.py`
    is touched (plus the new test file + baseline).
  - `LineageClustering.__version__` is bumped via #383's commit; this PR
    does not change it again.
  - The baseline npz is small (~25 KB) and committed as a fixture; the
    test gracefully `skipTest`s if absent and can be regenerated by
    `python -m tests.test_LineageClustering_equivalence regen`.
  - First-call JIT compile of `_build_lineage_for_event_kernel` is
    one-time per process; the `cache=True` flag persists across
    invocations.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,6 +21,10 @@ repos:
     rev: v1.7.7
     hooks:
     -   id: docformatter
+        # Pin to py3.12 — docformatter's transitive dep `untokenize 0.1.1`
+        # uses `ast.Constant.s` (removed in py3.14), so pre-commit.ci's
+        # default py3.14 env fails to build it.
+        language_version: python3.12
 
 -   repo: https://github.com/pre-commit/mirrors-mypy
     rev: v1.16.1
--- a/fuse/plugins/micro_physics/lineage_cluster.py
+++ b/fuse/plugins/micro_physics/lineage_cluster.py
@@ -26,12 +26,24 @@ NEST_NONE = (12, 0, 0)
 _TYPE_NAMES = ("alpha", "e-", "e+", "gamma", "neutron")
 _PARENTTYPE_NAMES = ("", "none", "neutron", "gamma")
 _CREAPROC_NAMES = (
-    "RadioactiveDecayBase", "compt", "conv", "phot",
-    "photonNuclear", "eBrem", "Transportation",
+    "RadioactiveDecayBase",
+    "compt",
+    "conv",
+    "phot",
+    "photonNuclear",
+    "eBrem",
+    "Transportation",
 )
 _EDPROC_NAMES = (
-    "RadioactiveDecayBase", "compt", "conv", "phot",
-    "Transportation", "hadElastic", "neutronInelastic", "nCapture", "ionIoni",
+    "RadioactiveDecayBase",
+    "compt",
+    "conv",
+    "phot",
+    "Transportation",
+    "hadElastic",
+    "neutronInelastic",
+    "nCapture",
+    "ionIoni",
 )
 
 _TYPE_CODE = {name: i for i, name in enumerate(_TYPE_NAMES)}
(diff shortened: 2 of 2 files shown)


HenningSE:
Hi @DonNabla I started looking at the PR and about half of the changes are conversions of strings into integers. I think with https://github.com/XENONnT/mc/pull/381 we would get similar numbers but on a lower level (directly when reading the root files) and this would make some of the changes here obsolete. I would prefer to not duplicate the work and finish https://github.com/XENONnT/mc/pull/381 first, then trim this PR to the required minimum.

SaelMo:
PR done @HenningSE

DonNabla:
I will look into it early next week

